PR #59, opened 2026-05-13
fix(login): Uses the form tenant-id at login

mariowise:
This PR changes the way the interceptor captures the login call so that the backoffice app can connect to backends with custom `tenant-id` header. Currently the value received in the form is ignored and the environment one is used.


Aman-Mittal (approved):
Seems OK to merge. Make sure to Link Ticket in future

PR #193, opened 2026-08-02
feat(products): configure interest recalculation and the remaining product settings

### LINKED ISSUE #192: Loan product form cannot enable interest recalculation, and hardcodes it to false
## Summary

Interest recalculation cannot be configured on the loan product form. `isInterestRecalculationEnabled`
is set to a hardcoded `false` in the create defaults and read back on edit, but there is no control
for it and none of its dependent settings exist in the form at all.

Four further product settings are also unexposed: `chargeOffBehaviour`,
`enableAccrualActivityPosting`, `fixedLength` and `repaymentStartDateType`.

## Business value

**A product created here can never use interest recalculation, and one that does is quietly at
risk when edited.**

- **Declining-balance products cannot be created properly.** Interest recalculation is what makes
  interest respond to what the borrower actually owes on the day, rather than to the original
  schedule. Without it, every product created in this application charges interest on the planned
  balance regardless of early or late payment. For an institution offering early-settlement
  benefits, that is the difference between the product it sells and the product the system runs.
- **The setting is write-once and unreachable.** Creating a product always sends `false`. Editing
  reads the existing value back, so a product configured elsewhere keeps its setting — but only by
  accident, and nothing in the form lets anyone see or change it.
- **Charge-off behaviour is not selectable.** `chargeOffBehaviour` decides what happens to interest
  when a loan is charged off — carry on as normal, stop accruing, or bring the maturity forward.
  That is an accounting-policy decision the institution cannot express here.
- **Accrual activity posting, fixed-length terms and the repayment start date rule** are likewise
  unavailable.

## The conditional rules are documented, unusually

`POST /loanproducts` states them explicitly, which most of this API does not:

> Additional Mandatory Fields if interest recalculation is enabled(true):
> `interestRecalculationCompoundingMethod`, `rescheduleStrategyMethod`,
> `recalculationRestFrequencyType`
>
> Additional Optional Fields if interest recalculation is enabled(true):
> `isArrearsBasedOnOriginalSchedule`, `preClosureInterestCalculationStrategy`
>
> … if `recalculationRestFrequencyType` is not same as repayment period:
> `recalculationRestFrequencyInterval`
>
> … if `interestRecalculationCompoundingMethod` is enabled: `recalculationCompoundingFrequencyType`
>
> … and if that is not same as repayment period: `recalculationCompoundingFrequencyInterval`

The form should follow that chain rather than showing every field at once: the mandatory three
appear with the toggle, and each further field appears only when the setting it depends on calls
for it.

## A correction to carry with this

#191 added the capitalised income and buy-down option lists as local constants, with a comment
saying the product template does not return them. **That was wrong.** The template returns
`capitalizedIncomeTypeOptions`, `capitalizedIncomeCalculationTypeOptions`,
`capitalizedIncomeStrategyOptions` and the three `buyDownFee*Options`, along with every option
list this issue needs — `rescheduleStrategyTypeOptions`,
`preClosureInterestCalculationStrategyOptions`, `repaymentStartDateTypeOptions`,
`chargeOffBehaviourOptions`, `interestRecalculationCompoundingTypeOptions` and
`interestRecalculationFrequencyTypeOptions`.

Those constants should be replaced with the template's lists, so labels come from the server and a
value added upstream appears without a code change.

## Scope

- Interest recalculation toggle and its dependent fields, following the documented rule chain.
- `chargeOffBehaviour`, `enableAccrualActivityPosting`, `fixedLength`, `repaymentStartDateType`.
- Replace the local option constants from #191 with the template's lists.
- Carry every new field through `loadProductData`, which rebuilds the payload field by field.

## Comprehensibility for non-specialists

"Compounding method", "rest frequency", "reschedule strategy" and "pre-closure interest
calculation" are specialist terms that mean little without context. Each needs help text saying
what the institution is choosing and what the borrower experiences, consistent with #187, #189
and #191.

## Acceptance criteria

- [ ] Interest recalculation is settable, and its mandatory fields appear with it.
- [ ] The conditionally-optional fields appear only when their condition holds.
- [ ] Turning it off clears the whole group rather than leaving values in the payload.
- [ ] The four remaining settings are selectable, driven by the template's option lists.
- [ ] The local option constants from #191 are gone.
- [ ] Existing values load on edit and survive a save round trip.
- [ ] Unit tests cover the rule chain and the clearing; a mocked Playwright spec asserts the
      submitted payload; the recorded demo shows recalculation being configured.
- [ ] Every new control carries help text that reads sensibly to a non-specialist.

Aman-Mittal:
Closes #192. Completes the product-configuration phase begun in #189 and #191.

Interest recalculation could not be configured at all. `isInterestRecalculationEnabled` was a
hardcoded `false` in the create defaults and read back on edit — so creating a product here always
disabled it, a product configured elsewhere kept its setting only by accident, and nothing let
anyone see or change it. Its seven dependent fields did not exist in the form.

### Following a rule chain the API actually documents

`POST /loanproducts` states the conditions explicitly, which is unusual for this client:

> Additional Mandatory Fields if interest recalculation is enabled(true):
> `interestRecalculationCompoundingMethod`, `rescheduleStrategyMethod`,
> `recalculationRestFrequencyType`
>
> … if `recalculationRestFrequencyType` is not same as repayment period:
> `recalculationRestFrequencyInterval`
>
> … if `interestRecalculationCompoundingMethod` is enabled: `recalculationCompoundingFrequencyType`
>
> … and if that is not same as repayment period: `recalculationCompoundingFrequencyInterval`

The form follows that chain rather than showing everything at once, and switching recalculation
off clears the whole group instead of leaving recalculation settings on a product that does not
recalculate. The conditions match on each option's **code**, not its id, so Fineract's numbering
doesn't become a magic number here.

Also adds `chargeOffBehaviour`, `enableAccrualActivityPosting`, `fixedLength` and
`repaymentStartDateType`.

### A correction to #191

That PR added the capitalised income and buy-down option lists as local constants, with a comment
saying the product template does not return them. **It does** — `capitalizedIncomeTypeOptions`,
`capitalizedIncomeCalculationTypeOptions`, `capitalizedIncomeStrategyOptions` and the three
`buyDownFee*Options`, along with every list this PR needed. I got that wrong; the constants are
gone and all six now come from the template, so labels are the server's and a value added upstream
appears without a code change.

### The object-in-a-signal trap, again

The conditional flags are held as their own signals rather than read off the payload. `product` is
a signal holding an object, so assigning to one of its properties invalidates nothing and a
`computed` reading `product().x` never re-runs. This is the third time this pattern has bitten
(#189 for the down payment controls, #191's checkboxes) — here the unit tests caught it before it
reached review, but it's a sharp edge in this component that would be worth removing properly at
some point.

### Edit round trip

Every new field is carried through `loadProductData`, which rebuilds the payload field by field.
The response nests recalculation under `interestRecalculationData` and names the compounding
frequency differently from the request (`interestRecalculationCompoundingFrequencyType` vs
`recalculationCompoundingFrequencyType`); the round trip handles both.

### Comprehensibility

The vocabulary here is the most specialist in the form, so the help does the most work:

> **Recalculate interest on the actual balance** — Charge interest on what the borrower actually
> owes on the day, rather than on the balance the original schedule assumed. A borrower who pays
> early then pays less interest, and one who pays late pays more.

Labels avoid jargon where plain words work: *What gets compounded*, *When the amount owed changes*,
*How often interest is recalculated*, *Interest on early settlement*, *When the loan is charged off*.

### Verification

| Check | Result |
|---|---|
| Unit tests | **729 passing** (720 → 729) |
| Mocked Playwright | **208/208 passing** (204 → 208) |
| `tsc` (app + spec) | 0 errors |
| `npm run build` | passes |
| lint (empty suppressions baseline), format, i18n, icons, license | all clean |

Driven by hand against mocks before committing: with recalculation on, the mandatory three appear
and neither interval does — correct, since the seeded frequency is "same as repayment period" and
the seeded compounding method is "none".

### Not in this PR

The demo step runs only in the `backend` project, which needs a live Fineract, so it is unverified
here; its assertions mirror the mocked spec that does run.

`isArrearsBasedOnOriginalSchedule`, the `*OnDayType` fields and the accounting mappings
(`capitalizedIncomeClassificationToIncomeAccountMappings`) remain unexposed — accounting mappings
are a separate area of the form, and the day-type fields only apply to frequency options this
deployment does not currently surface.

The pre-existing `NG0100` reports on this form (`ng-untouched`, and value changes on plain
`[(ngModel)]` fields) are unchanged — the form-state class described earlier, advisory rather than
enforced.

### DIFF
--- a/e2e/full-demo.spec.ts
+++ b/e2e/full-demo.spec.ts
@@ -200,6 +200,17 @@ test.describe('Full feature demo recording', () => {
     await page.getByRole('spinbutton', { name: INTEREST_RATE_LABEL }).fill('10');
     await page.getByRole('spinbutton', { name: NUMBER_OF_REPAYMENTS_LABEL }).fill('3');
     await page.getByRole('spinbutton', { name: REPAYMENT_EVERY_LABEL }).fill('1');
+
+    // Interest recalculation charges interest on what the borrower actually owes rather than on
+    // the planned balance. It applies to both engines, and its dependent fields follow the chain
+    // the API documents: the mandatory three appear with the toggle, and each interval appears
+    // only when its frequency is something other than the repayment period.
+    await page.getByTestId('loan-product-interest-recalculation').click();
+    await expect(page.getByTestId('loan-product-compounding-method')).toBeVisible();
+    await expect(page.getByTestId('loan-product-reschedule-strategy')).toBeVisible();
+    await expect(page.getByTestId('loan-product-rest-interval')).toHaveCount(0);
+    await beat(page);
+
     await page.getByRole('button', { name: 'Save' }).click();
     await expect(page).toHaveURL(/\/products\/loan$/, { timeout: 15000 });
 
--- a/e2e/full-demo.spec.ts
+++ b/e2e/full-demo.spec.ts
@@ -208,7 +208,18 @@ test.describe('Full feature demo recording', () => {
     await page.getByTestId('loan-product-interest-recalculation').click();
     await expect(page.getByTestId('loan-product-compounding-method')).toBeVisible();
     await expect(page.getByTestId('loan-product-reschedule-strategy')).toBeVisible();
+    // No interval: the seeded rest frequency is the repayment period, which needs none.
     await expect(page.getByTestId('loan-product-rest-interval')).toHaveCount(0);
+    // Fineract only supports recalculation with daily interest calculation, so the form fixes
+    // that field and says why rather than letting the server reject the product.
+    await expect(page.getByTestId('interest-calc-period-locked-note')).toBeVisible();
+    await beat(page);
+
+    // Turned back off before saving. This product is the one the rest of the walkthrough lends
+    // against, and a recalculating product constrains every loan created from it — the demo is
+    // here to show the controls, not to change the product ten later steps depend on.
+    await page.getByTestId('loan-product-interest-recalculation').click();
+    await expect(page.getByTestId('interest-calc-period-locked-note')).toHaveCount(0);
     await beat(page);
 
     await page.getByRole('button', { name: 'Save' }).click();
--- a/e2e/loan-product-down-payment.spec.ts
+++ b/e2e/loan-product-down-payment.spec.ts
@@ -370,6 +370,9 @@ test.describe('Loan product down payment and tranches', () => {
     // Seeded with "same as repayment period", which needs no interval.
     await expect(page.getByTestId('loan-product-rest-interval')).toHaveCount(0);
 
+    // Wait for the select to be attached before driving it: its options come from the template
+    // request, and picking from it before that resolves times out under parallel load.
+    await expect(page.getByTestId('loan-product-rest-frequency')).toBeVisible();
     await selectOption(page, 'How often interest is recalculated', 'Daily');
 
     await expect(page.getByTestId('loan-product-rest-interval')).toBeVisible();
--- a/e2e/loan-product-down-payment.spec.ts
+++ b/e2e/loan-product-down-payment.spec.ts
@@ -107,6 +107,50 @@ async function login(page: Page): Promise<Captured> {
         creditAllocationAllocationTypes: [],
         creditAllocationTransactionTypes: [],
         chargeOptions: [],
+        capitalizedIncomeTypeOptions: [
+          { code: 'FEE', value: 'Fee' },
+          { code: 'INTEREST', value: 'Interest' },
+        ],
+        capitalizedIncomeCalculationTypeOptions: [{ code: 'FLAT', value: 'Flat' }],
+        capitalizedIncomeStrategyOptions: [
+          { code: 'EQUAL_AMORTIZATION', value: 'Equal amortization' },
+        ],
+        buyDownFeeIncomeTypeOptions: [{ code: 'FEE', value: 'Fee' }],
+        buyDownFeeCalculationTypeOptions: [{ code: 'FLAT', value: 'Flat' }],
+        buyDownFeeStrategyOptions: [{ code: 'EQUAL_AMORTIZATION', value: 'Equal amortization' }],
+        interestRecalculationCompoundingTypeOptions: [
+          { id: 0, code: 'interestRecalculationCompoundingMethod.none', description: 'None' },
+          { id: 1, code: 'interestRecalculationCompoundingMethod.fee', description: 'Fee' },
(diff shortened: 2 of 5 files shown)


Aman-Mittal:
Pushed `930ed58`, which fixes a real bug in the previous commit that the backend e2e project caught.

### What was wrong

Fineract only supports interest recalculation alongside **daily** interest calculation. Anything
else is rejected with `not.supported.for.selected.interest.calculation.type` — and this form's own
default, "same as repayment period", is one of the rejected values.

So the control this PR adds produced a product the server refused: the user ticked the box, pressed
Save, and stayed on the create page with a raw validation error.

I confirmed it against a live Fineract before changing anything, rather than inferring it from the
failure:

```
interestCalculationPeriodType: 1  + recalculation → 400 not.supported.for.selected.interest.calculation.type
interestCalculationPeriodType: 0  + recalculation → accepted
```

### The fix

The form sets daily when recalculation is switched on, locks the control and explains why — the
same pattern as the repayment strategy lock on progressive products:

> Fixed to Daily because interest recalculation is on — Fineract only supports the two together.
> Turn recalculation off to choose a different period.

Two unit tests cover it: that enabling recalculation moves the period from `1` to `0`, and that the
explanation renders.

### The demo step was also wrong

It left recalculation enabled on the cumulative product that the following ten steps lend against.
A recalculating product constrains every loan created from it, so loan creation failed a few steps
later — which is why the CI failure moved down the file rather than disappearing when I fixed the
save.

The step now shows the controls, including the locked note, and turns recalculation back off before
saving. The demo is there to show the UI, not to change the product the rest of the walkthrough
depends on.

### Verification

I ran the **backend** project against a local Fineract this time, which is what I could not do when
the PR was opened:

| Check | Result |
|---|---|
| Backend Playwright (incl. the failing demo) | **16/16 passing** |
| Mocked Playwright | **208/208 passing** |
| Unit tests | **731 passing** |
| `tsc`, build, lint, format, i18n, licence | all clean |

One mocked test flaked once under three parallel workers and passed in isolation and on re-run; I
added an explicit wait for the select to attach before driving it rather than leave that to CI's
retries.

### Note

This is the second thing in this phase that only a live backend would have caught, the first being
the missing `IonCheckbox` import in #189. The mocked specs assert what the UI sends; they cannot
assert that Fineract accepts it. Worth considering whether product-creation paths should have at
least one backend-project test each, rather than relying on the demo to exercise them incidentally.

Issue #216, opened 2026-08-02
Deployed nginx config sets no HTTP security headers


Aman-Mittal:
Part of the GA security hardening work.

## Summary

`deploy/nginx.conf` — the config the shipped container serves the application with — sets no
security headers at all. A grep for `Content-Security-Policy`, `Strict-Transport-Security`,
`X-Frame-Options`, `X-Content-Type-Options` and `Referrer-Policy` returns zero matches.

## Business value

**These are the backstop for everything the application cannot control.**

The codebase is in good shape on the application side: no `bypassSecurityTrust`, no `innerHTML`
sinks, no raw HTML rendering, and the GA check gates that so a single careless commit cannot
remove it. But that only covers what this code does. Headers cover what a browser will permit
regardless:

- **Without a CSP**, any injected string that does reach the DOM can load and execute remote
  script, and can exfiltrate to any host. With one, the same injection has nowhere to send data.
- **Without `X-Frame-Options` / `frame-ancestors`**, the application can be framed and
  clickjacked — a real risk for a back office where a single click approves or disburses money.
- **Without `nosniff`**, a response the server mislabels can be reinterpreted as script.
- **Without `Referrer-Policy`**, account and client identifiers in the URL leak to any external
  host the user navigates to.
- **Without HSTS**, a first request over plain HTTP is interceptable.

A back office that moves money is exactly the deployment where these are expected, and their
absence is the kind of thing a security review finds immediately.

## Scope

Set all five on the server block, with `always` so they are present on error responses too — which
is where a reflected payload would land.

The CSP should not admit `unsafe-eval`: the AOT build does not need it, and it would give an
injected string a route to execution. `unsafe-inline` is needed for styles only, because Angular
emits component styles inline and there is no nonce plumbed through a static file server; scripts
should not get it.

`connect-src` should allow the same origin only. A deployment whose Fineract is on another host
adds that origin here as well as to the application's allow-list — the two are deliberately
separate, so a browser-side setting alone cannot open a new destination.

## Acceptance criteria

- [ ] All five headers set, with `always`.
- [ ] The config passes `nginx -t`.
- [ ] The headers are observable on a real response from the container image.
- [ ] `npm run ga:check` reports this gate as passing.

Aman-Mittal:
Done and merged in #217.

`deploy/nginx.conf` now sets all five headers on the server block, each with `always` so they
survive error responses — which is where a reflected payload would land.

[...]

Issue #215, opened 2026-08-02
Login form offers third-party API hosts as one-click options


Aman-Mittal:
Part of the GA security hardening work.

## Summary

The server picker on the login form hard-codes external demo endpoints as selectable options,
alongside the deployment's own. They ship in every production build.

## Business value

**A credential-entry form should not invite anyone to post credentials to someone else's server.**

A teller who picks the wrong entry from a dropdown types their real username and password into a
third-party host. There is no warning, and the option sits next to the legitimate one in the same
list. Even setting aside malice, this is a routine way to leak working credentials — a
mis-click during training, or a stale selection remembered from a previous session.

It also makes the unvalidated endpoint override materially worse: the dropdown is the discoverable
path to it.

## Scope

Offer only what the deployment sanctions: its own configured endpoint, the same-origin proxy, and
any origins named in `config.json`. A custom URL remains possible for operators, but is checked
against the same allow-list as everything else rather than accepted on trust.

## Acceptance criteria

- [ ] No external hostnames remain hard-coded in the component.
- [ ] The picker is built from deployment configuration.
- [ ] A custom URL is validated before any credential is sent.
- [ ] `npm run ga:check` reports this gate as passing.

Aman-Mittal:
Done and merged in #217.

The hard-coded external hosts are gone from `login.component.ts`. The server picker is now built
from deployment configuration — the configured endpoint, the same-origin proxy, and any origins
named in `allowedApiOrigins`. A custom URL is still possible for operators, but is checked against
the same allow-list rather than accepted on trust, and the form explains the refusal instead of
failing silently.

`npm run ga:check` reports this gate as passing: no hostname literals remain in the component.

Issue #214, opened 2026-08-02
API endpoint override accepts any host, including one an attacker chooses


Aman-Mittal:
Part of the GA security hardening work.

## Summary

`ConfigService.setApiUrl()` writes whatever it is given straight into the running config and into
local storage:

```ts
setApiUrl(url: string): void {
  this._config.update((config) => ({ ...config, fineractApiUrl: url }));
  this.storage.write('runtimeConfig', { fineractApiUrl: url });
}
```

There is no validation. The stored value is also read back on the next load and applied over the
deployment's own `config.json`.

## Business value

**Whatever this names receives the user's credentials on their next sign-in.**

The override exists for a good reason — an operator pointing the application at their own Fineract
— but as written it is a general redirect. Two routes to abuse:

- Anything running as the page can call it, or write the local-storage key directly, and the
  choice survives a reload. Combined with the unrestricted `Authorization` header, the next login
  posts real credentials to the attacker's host.
- A user can be talked into pasting a URL into the endpoint field. It looks like configuration,
  not like a credential handover.

Local storage is writable by anything running as the page, so a stored override cannot be treated
as more trustworthy than a fresh one.

## Scope

Validate against an allow-list. Same-origin endpoints are always permitted; an absolute one must
match this document's origin or an origin the deployment named in `config.json`. Refuse anything
else, both when set and when read back from storage. The setter reports refusal so callers can
surface it rather than silently continuing.

## Acceptance criteria

- [ ] A same-origin endpoint is accepted and persisted.
- [ ] An absolute endpoint not on the allow-list is refused, and nothing is persisted.
- [ ] An absolute endpoint the deployment allow-listed is accepted.
- [ ] A stored endpoint that is not allow-listed is ignored on load.
- [ ] The login form refuses before authenticating, so the password never reaches an unsanctioned
      host, and explains why.
- [ ] Unit tests cover each case.

Aman-Mittal:
Done and merged in #217.

`setApiUrl` now validates against an allow-list and returns whether it accepted, so the login form
refuses **before** authenticating rather than after — the password never reaches an unsanctioned
host. Same-origin endpoints are always permitted; an absolute one must match this document's
origin or an origin the deployment named in `config.json` via the new `allowedApiOrigins` key.

A stored override clears the same bar on load: local storage is writable by anything running as
the page, so it is not more trusted than a fresh value.

One bug worth recording, caught while implementing: the check initially read the config signal,
[...]

Issue #213, opened 2026-08-02
Authorization header is sent to every destination, not just the API


Aman-Mittal:
Part of the GA security hardening work.

## Summary

`authInterceptor` attaches `Authorization: Basic <token>` to every outgoing `HttpClient` request
whenever a session exists. There is no check on the destination — confirmed by reading the file:
no URL comparison appears anywhere in it. The tenant header goes out unconditionally too.

## Business value

**The first third-party integration ships banking credentials to a vendor, and nothing reports
it.**

Today nothing in the application calls a third party, so there is no live leak. That is exactly
what makes this cheap to fix now and expensive later: the moment someone adds an analytics pixel,
an error reporter, a document-storage SDK or a credit-bureau call through the same `HttpClient`,
that vendor starts receiving credentials that unlock the institution's entire Fineract instance.
Nobody would see it happen — the header is added by an interceptor nobody edits, in a file nobody
revisits.

The tenant header matters as well, if less severely: it names the institution, so sending it
off-origin discloses which organisation the deployment belongs to.

## Scope

Attach credentials and tenant headers only when the request is bound for the configured API. A
relative URL is same-origin by construction and stays trusted; an absolute one is compared by
origin, so a path change on the same host remains trusted and a different host does not. A
deployment whose Fineract is legitimately on another origin is supported — that one origin is
trusted, and only that one.

## Acceptance criteria

- [ ] No `Authorization` header on a request to a foreign origin.
- [ ] No tenant header on such a request either.
- [ ] Relative and same-origin absolute requests are unaffected.
- [ ] A cross-origin API configured by the deployment still authenticates.
- [ ] Unit tests cover each of the above.

Aman-Mittal:
Done and merged in #217.

`authInterceptor` now attaches credentials and the tenant header only when the request is bound
for the configured API. A relative URL is same-origin by construction; an absolute one is compared
by origin, so a cross-origin API a deployment configured still authenticates, and nothing else
does. The tenant header is withheld from foreign origins too, since it names the institution.

Four unit tests cover it: no `Authorization` and no tenant header to a foreign origin, and both
still present for relative and same-origin absolute requests.

`npm run ga:check` reports this gate as passing.

PR #217, opened 2026-08-02
fix(security): close the four blocking GA readiness gates

### LINKED ISSUE #212: GA security hardening: close the four blocking readiness gates
Tracking issue for the blocking failures reported by `npm run ga:check`.

At the time of writing the check reported **3/8 gates passing, 4 blocking failures**. Each was
verified in the code rather than taken on the report's word.

## The failures, and why they compound

1. **`Authorization` is sent to every destination.** `authInterceptor` clones in `Basic <token>`
   whenever a session exists, with no check on where the request is going.
2. **`setApiUrl()` accepts any string.** It writes straight to config and local storage with no
   validation, so the endpoint the credentials are posted to can be set to anything.
3. **The login form offers third-party API hosts.** External demo endpoints were hard-coded as
   selectable options on a credential-entry form.
4. **`deploy/nginx.conf` sets no security headers.** No CSP, HSTS, `X-Frame-Options`,
   `X-Content-Type-Options` or `Referrer-Policy`.

These are worse together than apart. (3) makes (2) a one-click action, and (2) makes (1) a
credential leak: point the app at a host you control, and the next request carries the user's
Basic credentials to it. (4) removes the browser-side backstop that would otherwise limit where
an injected script could send anything.

Nothing in the application calls a third party today, which is precisely why this is cheap now —
the first analytics pixel, error reporter or vendor SDK sharing this `HttpClient` would inherit
the problem silently.

## Sub-issues

- Restrict the `Authorization` header to the API origin
- Validate the API endpoint override against an allow-list
- Remove hard-coded third-party hosts from the login form
- Set HTTP security headers in the deployed nginx config

## Not in scope

The advisory adapter-boundary backlog is a long migration tracked separately; `npm run lint`
already blocks new violations, so the count only falls.

Aman-Mittal:
Closes #212, #213, #214, #215, #216.

`npm run ga:check` reported **3/8 gates passing, 4 blocking failures**. This takes it to **7/8,
0 blocking** — the remaining one is the advisory adapter-boundary backlog.

I confirmed each failure in the code before changing anything, rather than taking the report's
word for it.

### Why these were worse together than apart

The login form offered third-party hosts as one-click options. The endpoint override accepted any
string and persisted it. The auth interceptor attached `Authorization: Basic` to **every** outgoing
request regardless of destination — no URL comparison appeared anywhere in the file.

Chain them and you have a supported path to pointing the application at an attacker's host and
posting real credentials to it on the next sign-in. Missing security headers removed the
browser-side backstop that would otherwise limit where an injected script could send anything.

Nothing here calls a third party today, which is exactly why this was cheap now: the first
analytics pixel, error reporter or vendor SDK sharing this `HttpClient` would have inherited the
problem silently.

### What changed

- **Interceptor** sends credentials — and the tenant header — only to the configured API. Relative
  URLs are same-origin by construction; absolute ones are compared by origin, so a cross-origin
  API the deployment configured still authenticates and nothing else does. The tenant header is
  withheld too, since it names the institution.
- **`setApiUrl`** validates against an allow-list and *returns* whether it accepted, so the login
  form stops **before** authenticating rather than after. A stored override clears the same bar:
  local storage is writable by anything running as the page, so it is not more trusted than a
  fresh value.
- **Server picker** is built from deployment configuration instead of hard-coded hosts.
- **`deploy/nginx.conf`** sets CSP, HSTS, `X-Frame-Options`, `X-Content-Type-Options` and
  `Referrer-Policy`, all with `always` so they survive error responses — which is where a
  reflected payload would land. The CSP admits no `unsafe-eval`; `unsafe-inline` is styles-only.

### A bug I caught while writing it

The allow-list check initially read the config signal. But `getStoredOverride` runs inside a
**field initialiser**, before that signal exists — it would have thrown on startup, in the one code
path every user hits. The predicate now takes the allowed origins as an argument, and the loader
passes the freshly-loaded list rather than reading back a config that has not been merged yet.

### Verification

| Check | Result |
|---|---|
| `npm run ga:check` | **3/8 → 7/8**, 0 blocking |
| Unit tests | **766 passing** (744 → 766) |
| Mocked Playwright | **210/210 passing** |
| `nginx -t` on the real image | passes |
| Headers observed on a live response | all five present |
| `tsc` (app + spec), build, lint, format, i18n, icons, licence | clean |

Headers were checked by serving the config from the actual `nginx:alpine` image and reading the
response, not by grepping the file.

### Two notes for review

The existing `ConfigService` specs asserted the **old permissive behaviour** — they set
cross-origin URLs and expected them to stick. Those now assert the new rule, with added cases for
refusal and for a non-allow-listed stored value being ignored.

Storage access in that spec is funnelled through three helpers, which takes the adapter-boundary
suppression baseline **down rather than up: 435 → 431**. New violations would have failed lint;
adding suppressions would have been the wrong direction.

### DIFF
--- a/deploy/nginx.conf
+++ b/deploy/nginx.conf
@@ -22,6 +22,26 @@ server {
     root /usr/share/nginx/html;
     index index.html;
 
+    # Security headers.
+    #
+    # `always` matters: without it nginx omits these on error responses, which is exactly where
+    # a reflected payload would land.
+    #
+    # The CSP has no 'unsafe-eval'. Angular's AOT build does not need it, and admitting it would
+    # give any injected string a route to execution. 'unsafe-inline' remains for styles only:
+    # Angular emits component styles inline, and there is no nonce plumbed through this static
+    # server. Scripts do not get it.
+    #
+    # connect-src allows the same origin only. A deployment whose Fineract lives on another host
+    # must add that origin here as well as to `allowedApiOrigins` in config.json — the two are
+    # deliberately separate, so a browser-side setting alone cannot open a new destination.
+    add_header Content-Security-Policy "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; font-src 'self' data:; connect-src 'self'; frame-ancestors 'none'; base-uri 'self'; form-action 'self'; object-src 'none'" always;
+    add_header X-Content-Type-Options "nosniff" always;
+    add_header X-Frame-Options "DENY" always;
+    add_header Referrer-Policy "no-referrer" always;
+    # Only honoured over HTTPS; harmless on the plain-HTTP listener and correct once terminated.
+    add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
+
     # Gzip Compression
     gzip on;
     gzip_http_version 1.1;
--- a/eslint-suppressions.json
+++ b/eslint-suppressions.json
@@ -24,11 +24,6 @@
       "count": 4
     }
   },
-  "src/app/core/services/config.service.spec.ts": {
-    "no-restricted-globals": {
-      "count": 4
-    }
-  },
   "src/app/core/services/institution-config.service.spec.ts": {
     "no-restricted-globals": {
       "count": 5
--- a/src/app/core/interceptors/auth.interceptor.spec.ts
+++ b/src/app/core/interceptors/auth.interceptor.spec.ts
@@ -23,11 +23,14 @@ import { authInterceptor } from './auth.interceptor';
 import { AuthService } from '../services/auth.service';
 import { signal } from '@angular/core';
 import { of } from 'rxjs';
+import { provideTestConfig } from '../../testing/config';
 
 describe('authInterceptor', () => {
   let authServiceSpy: jasmine.SpyObj<AuthService>;
   const testUrl = '/test';
   const testTenant = 'test-tenant';
+  const foreignUrl = 'https://vendor.example.com/collect';
+  const TENANT_HEADER = 'Fineract-Platform-TenantId';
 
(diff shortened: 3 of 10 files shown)


Aman-Mittal:
### Secret-scanner alert on `3d55313` — resolved

A scanner flagged a "Generic High Entropy Secret" in my commit. It was **not a real credential**,
but the finding was legitimate and worth removing rather than dismissing.

**What it was.** A test fixture I added to `auth.interceptor.spec.ts`:

```ts
const token = 'dGVzdDp0b2tlbg==';   // base64 of "test:token"
```

Base64-of-`user:pass` is exactly the shape of a Basic credential, so a scanner has no way to tell
it from the real thing — and neither does a reviewer skimming the diff. That is the scanner working
correctly.

**What I changed.**

```ts
// Deliberately not base64-shaped. A realistic-looking encoded credential here trips secret
// scanners, and the assertions only care that the header echoes whatever the service returned.
const token = 'fake-token-for-tests';
```

The tests only assert the header echoes whatever `getAuthToken()` returned, so the encoding was
never doing any work.

**Remediation.** I amended rather than adding a follow-up commit, and force-pushed
(`3d55313` → `e987d7e`), so the string is gone from this branch's history rather than merely
corrected at the tip. Verified with `git log -S` and a grep across every commit in the branch —
both return zero.

I also checked the rest of the branch for anything else base64-shaped that I introduced: none.

766 unit tests still pass and lint is clean. **Please re-run the scanner against `e987d7e`.**

Worth noting for the repo generally: `base64EncodedAuthenticationKey: 'YmFzZTY0'` appears in
several pre-existing mocked e2e specs. Those pre-date this branch and I have not touched them, but
they are the same shape and may trip the same rule in future.

Issue #228, opened 2026-08-02
Adopt Apache Magpie for agent-assisted triage and maintenance


Aman-Mittal:
## Requirement

Adopt [Apache Magpie](https://magpie.apache.org/) ([apache/magpie](https://github.com/apache/magpie)) in this repository. Magpie became an ASF Top-Level Project on 29 June 2026. It is a curated set of agent recipes for the repetitive parts of running a project — issue and PR triage, contributor mentoring, fix drafting, security-report handling — with human review kept in the loop throughout.

This is a requirement for the project, not an open question. What remains open is *scope and sequencing*, tracked in the checklist below.

**Status: implementation open in #241**, pinning the signed **0.1.0** release with project overrides.

## Business value

This repository is young, moving quickly, and has a small number of people doing the maintenance. The cost of that shows up in specific, recurring places:

- **Triage backlog.** Issues arrive faster than they get labelled, scoped and routed. Newcomer-sized work in particular tends to sit unlabelled, which is exactly the work that most needs to be findable.
- **Newcomer onboarding.** Getting a first-time contributor productive here means explaining an Angular 22 signal-based codebase, an adapter boundary, a generated API client that must not be hand-edited, and an e2e suite split between mocked and real-backend projects. That explanation is currently repeated by hand each time.
- **Review load.** Reviewers spend attention on mechanical checks — conventions, missing tests, missing translations — before reaching the parts that actually need judgement.

Magpie targets each of these directly, and it does so without adding a vendor dependency: it is explicitly vendor-neutral across models, and its stated design constraints are privacy, security and supply-chain integrity, with an affordability commitment so participation does not require a paid subscription.

The thing it buys is maintainer attention. Every hour not spent labelling and re-explaining is an hour available for design and review, which is the scarce resource here.

## Non-negotiables for the adoption

These are requirements on *how* it is adopted, and any implementation has to satisfy them:

1. **Vendor and model neutral.** Skills install to `.agents/skills/`, the canonical path shared by Codex, Cursor, Gemini CLI, Copilot, OpenCode, Cline, Zed and Warp, with per-tool directories as relay symlinks. No model vendor, runtime, or provider-specific feature may be named in any committed override. Anything requiring a paid subscription belongs in the gitignored personal override directory, so participation never depends on one contributor's account.
[...]

PR #241, opened 2026-08-03
Adopt Apache Magpie 0.1.0 with project overrides

Aman-Mittal:
Implements #228.

Adopts [Apache Magpie](https://magpie.apache.org/) 0.1.0. Nothing runs on its own — no workflow, no schedule, no bot. A maintainer invokes a skill, reviews the output, and decides.

## Verified supply chain

Installed from the **signed ASF release** rather than a `git clone`:

```
sha512sum -c  → apache-magpie-0.1.0-source.zip: OK
gpg --verify  → Good signature from "Jarek Potiuk (apache) <potiuk@apache.org>"
```

The signing key is published in the project's own `KEYS` file. `.apache-magpie.lock` records the SHA-512, so any later re-fetch is verified against the same value, and a change to `url`/`ref`/`sha512` is reviewed as a supply-chain change.

> The upstream install recipes are stale on two points: they describe the released-zip method as *"not yet available; use Method 3"* and name the artefact `-source-release.zip`. 0.1.0 shipped on 2026-08-03 as `apache-magpie-0.1.0-source.zip`. Worth an upstream doc PR.

## Committed footprint

| Path | Tracked | Purpose |
|---|---|---|
| `.apache-magpie.lock` | yes | The version pin |
| `.agents/skills/magpie-setup/` | yes | Bootstrap, so a fresh clone can restore the rest |
| `.claude/skills/…`, `.github/skills/…` | yes | Relay **symlinks** (mode `120000`), not copies |
| `.apache-magpie-overrides/` | yes | This project's instructions to the skills |
| `.apache-magpie/` | **no** | Framework snapshot — a build artefact |

The framework never enters the repository's history. Contributors who do not use it are unaffected: no `npm` dependency, no build step, nothing to install.

## Vendor and model neutrality

Structural, not aspirational. Skills live in `.agents/skills/` — the canonical path read by Codex, Cursor, Gemini CLI, Copilot, OpenCode, Cline, Zed and Warp — with per-tool directories as relay symlinks into it, so there is no separate installation per tool and no default runtime.

No model vendor, provider or runtime is named anywhere in the committed overrides; they are written against capabilities, not any client's tool names or output format. Anything requiring a paid subscription belongs in the gitignored `.apache-magpie-local/`, so participation never depends on one contributor's account.

## The overrides are the substance

Generic recipes would ask for things this project deliberately does not do — explicit `OnPush` (already the v22 default), a facade over the generated client (rejected in ADR-0001), assigning issues (committer-only here) — and, more importantly, would **trust this repository's own checks**.

That trust is not warranted. Three separate mechanisms have reported success while testing nothing: `npm run test` ran 2 specs instead of 775 and exited `0`; Karma prints `TOTAL: n SUCCESS` after a bundle failure; the API drift manifest matched only single-line calls and quietly covered less on each regeneration.

So every override asks for **evidence that would fail if the claim were false**.

| Override | Encodes |
|---|---|
| `pr-management-code-review` | Four upgradability axes; OnPush-is-already-default; the object-in-a-signal trap; the adapter boundary and shrink-only suppressions baseline; wire-format rules that fail at runtime and never at compile time |
| `workflow-security-audit` | ASF Actions policy (only `apache/*`, `github/*`, `actions/*` unreviewed; SHA pins otherwise); untrusted-input paths; build-time-authored `config.json` as a credential-routing decision |
| `dependency-audit` | Generated-client carve-out; the scheduled spec sync as a supply-chain input; the positional-argument hazard it creates; runtime vs build-time exposure |
| `issue-triage` | Label vocabulary; the Business Value requirement; symptom-vs-cause evidence bar; never assign |
| `good-first-issue-author` | Scoping so nothing blocks; verified worked example; not-a-good-first-issue if it needs a live platform to verify |
| `pairing-self-review` | Check the count, not the word; verify wire behaviour against a running platform |

Concrete example of what the code-review override catches — all of these compiled cleanly, passed lint and passed the unit suite:

- an unpadded day returning **500** rather than a validation error;
- an empty string rejected as a *value* where the field was meant to be omitted;
- a list endpoint answering **204** for every query, so the screen was always empty;
- a summary endpoint answering **200** with all totals zeroed because a query parameter was missing.

The security override also captures that `allowedApiOrigins` entries are offered to users as **selectable login endpoints**, so whatever they name receives the user's credentials — widening that list is a security change, not configuration.

## Repository hygiene

- ASF headers on every new file — `scripts/check-license.sh` passes.
- Vendored upstream source excluded from **RAT** and **Prettier**; reformatting it would create diff noise against every framework upgrade.
- `npm run format:check` and `npm run lint` clean. No application code is touched, so build and test behaviour is unchanged.

## Scope enabled

`setup` + `utilities` (always on), `issue`, `pr-management`, `repo-health`.

Deliberately not enabled: `security` (needs a PMC decision — it handles reports before they are public), `release-management` (this project does not cut releases on that cadence), `mentoring` / `contributor-growth` (revisit with steadier newcomer traffic). `pairing` has an override ready and can be enabled per-contributor locally.

`repo-health` is the intended first trial: this repository already runs Apache RAT, licence compliance, dependency integrity, `zizmor` and CodeQL, so its audits should largely agree with checks already trusted — a low-risk way to confirm the framework behaves sensibly here before leaning on it for triage.


DOCS/MAGPIE.md
```diff
@@ -0,0 +1,123 @@
...
+
+See `.apache-magpie-overrides/README.md` for the index. The theme running through all of them
+is that **a passing check in this repository is weaker evidence than it looks** — three
+separate mechanisms have reported success while testing nothing — so every override asks for
+evidence that would fail if the claim were false.
+
+## Upgrading
+
+```bash
+/magpie-setup upgrade      # re-pins .apache-magpie.lock; commit the change
```
zhangxinyao88, line 102: This command does not seem to re-pin the committed lock. `upgrade.md` says `.apache-magpie.lock` is updated only when `bump-committed` is passed; plain `upgrade` just refreshes the local snapshot from the existing pin. Could we use the actual lock-bump command here and describe plain `upgrade` as a local sync?

.agents/skills/magpie-setup/adopt.md
```diff
@@ -0,0 +1,1537 @@
...
+
+   A fresh clone needs the snapshot populated before any
+   framework skill is invocable. In your agent harness, run:
+
+       /magpie-setup
+
+   (or follow [`.agents/skills/magpie-setup/`](.agents/skills/magpie-setup/))
+   to fetch the snapshot per the committed lock, scaffold the
+   gitignored symlinks, and install the post-checkout hook
+   that re-creates them on each worktree checkout.
```
zhangxinyao88, line 1224: This conflicts with the explanation above. Lines 1158-1171 say symlink reconciliation was removed from the hook and now happens lazily when the agent runs `verify`. Could this template describe the current behavior and mention the step needed to initialize a new worktree?

.agents/skills/magpie-setup/adopt.md
```diff
@@ -0,0 +1,1537 @@
...
+the broken line should remove it; the
+[`setup-isolated-setup-update`](../setup-isolated-setup-update/SKILL.md)
+drift check surfaces stale hook content on a routine sweep.
+
+## Step 11 — Project doc updates (FRESH only)
+
+Update two adopter-facing docs so contributors discover the
+framework before they hit a "skill not found" error:
+
+1. **`README.md` (contributor-facing summary, REQUIRED if
```
zhangxinyao88, line 1182: This repository already has both `README.md` and `AGENTS.md`, but this PR does not update either one. Since this section marks the README entry as required and also calls for an AGENTS.md entry when that file exists, could we add short links to `DOCS/MAGPIE.md` from both files?

.agents/skills/magpie-setup/adopt.md
```diff
@@ -0,0 +1,1537 @@
...
+   or not (see [`tools/gmail/oauth-draft/README.md` → MCP server](../../tools/gmail/oauth-draft/README.md#mcp-server)).
+3. **Reflect the outcome** in the recommended permission allow-list
+   (add `mcp__gmail-plaintext__create_draft` alongside the Gmail read
+   tools — see [`verify.md`](verify.md) check 8d). It only ever creates
+   **unsent** drafts the human reviews and sends, so allow-listing it
+   avoids a prompt on every draft.
+
+## Step 10 — Worktree-aware post-checkout hook (FRESH only)
+
+Install `<repo-root>/.git/hooks/post-checkout` — a best-effort
```
zhangxinyao88, line 1076: The PR says nothing runs on its own, but after setup this hook runs on every checkout and worktree creation. Could the overview make that distinction clear? This section should also say what happens if the user already has a `post-checkout` hook, so setup does not replace unrelated hook logic.

.agents/skills/magpie-setup/adopt.md
Aman-Mittal: Yes you are right, but i am considering to scrap magie entirely for now.

PR #247, opened 2026-08-06
ci: cancel superseded workflow runs on pull requests

### LINKED ISSUE #235: Superseded CI runs are never cancelled, so a new push queues behind the old one
## What is wrong

None of the five workflows (`ci.yml`, `e2e.yml`, `codeql.yml`, `zizmor.yml`, `api-spec-sync.yml`) declare a `concurrency` group. Push twice to the same PR and both runs execute to completion, even though nobody will ever read the first result.

`ci.yml` alone has 16 jobs and `e2e.yml` runs a real Fineract, so a superseded run is not cheap.

## Business value

Two things get worse without this, and both hit hardest when iterating quickly — which is exactly when a fast signal matters most.

**Feedback gets slower.** ASF projects share a runner pool. Obsolete runs sit in that queue holding slots, so the run you actually care about starts later. Fix a lint error, push, and you wait behind the run for the code you just replaced.

**Results get ambiguous.** A PR briefly shows results from two different commits. The failure you are looking at may belong to a commit that no longer exists, which is a confusing thing to debug and an easy way to waste an afternoon.

## How to fix

Add a `concurrency` block at the top level of each workflow, after `on:`:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Two details matter:

- **Only cancel on pull requests.** On `main`, each commit's result is worth keeping — cancelling there loses history and can leave a commit with no recorded status. The expression above cancels for PRs and lets pushes to `main` run to completion.
- **`api-spec-sync.yml` should not cancel at all.** It is a scheduled job that opens a PR; interrupting it halfway is not useful. Either leave it alone or give it a group with `cancel-in-progress: false` so two scheduled runs cannot overlap.

Including `github.workflow` in the group keeps workflows from cancelling each other.

## ASF compliance

`concurrency` is native GitHub Actions syntax — no new action, so nothing to review against the ASF third-party Actions allowlist. (On `apache/*` repositories, anything outside `apache/*`, `github/*` and `actions/*` is blocked by default and must be INFRA-reviewed and SHA-pinned. This change adds no dependency at all.)

## Verifying

Push twice in quick succession to a PR branch and confirm the first run shows as cancelled while the second proceeds. Then confirm a push to `main` is **not** cancelled by a subsequent one.

## Picking this up

This touches only `.github/workflows/`, so it does not conflict with any of the source-code issues currently open.

NkwaTambe:
## What and why

Add a `concurrency` group to `ci.yml`, `e2e.yml` and `codeql.yml` so a new push to a pull request cancels the run for the code it replaced. This speeds up feedback (superseded runs no longer hold runner slots) and removes ambiguous results where a PR briefly shows two commits' worth of statuses.

The group is keyed on `${{ github.workflow }}-${{ github.ref }}` so the workflows never cancel each other, and `cancel-in-progress` is `${{ github.event_name == 'pull_request' }}` so only PR runs are cancelled — pushes to `main` still run to completion and each commit keeps its result.

Closes #235

## Verification

- Confirmed `zizmor.yml` and `api-spec-sync.yml` already declare `concurrency` groups, so only the three missing workflows were changed.
- All three edited files parse as valid YAML and pass `prettier --check`.
- This is native GitHub Actions syntax — no new action, so nothing to review against the ASF third-party Actions allowlist.

## Screenshots

Not applicable (non-UI change).

## Checklist

- [x] New component or service code uses the adapter boundary in `src/app/core/adapters/` instead of direct browser globals or imperative third-party APIs. (N/A — workflow-only change)
- [x] User-facing strings use translation keys. (N/A — no UI strings)
- [x] I added or updated tests appropriate to this change, or explained why tests were not needed. (Workflow-only change; verified YAML parses and passes Prettier. The repo's own `zizmor` and CodeQL `actions` analysis will lint the new blocks on CI.)
- [x] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. (N/A — no UI workflow changes)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,10 @@ on:
       - main
       - develop
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   dependencies:
     name: Dependency Integrity
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -25,6 +25,10 @@ on:
   schedule:
     - cron: '40 11 * * 4'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   analyze:
     name: Analyze (${{ matrix.language }})
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -24,6 +24,10 @@ on:
       - develop
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 # Granted per job rather than here: a workflow-level `pull-requests: write` hands
 # that token to every step in every job, including the ones that run npm install
 # and Playwright. Each job below re-declares the minimum it needs.


Aman-Mittal (approved):
LGTM

PR #260, opened 2026-08-08
fix: show domain rule errors for 403 responses

### LINKED ISSUE #258: Business-rule refusals are reported as &#34;you do not have permission&#34;, because Fineract returns 403 for them
## Business value

Fineract uses HTTP **403** for two unrelated things: you lack the permission, and *the operation is not allowed in this state*. `errorInterceptor` treats every 403 as the first, so the platform's explanation of the second is thrown away and replaced with a message that sends the user to the wrong place entirely.

Closing a group that still has active members answers:

```json
{
  "httpStatusCode": "403",
  "userMessageGlobalisationCode": "validation.msg.domain.rule.violation",
  "defaultUserMessage": "Errors contain reason for domain rule violation.",
  "errors": [
    {
      "defaultUserMessage": "Group cannot be closed because of active clients associated with it.",
      "userMessageGlobalisationCode": "error.msg.Group.close.active.clients.exist",
      "parameterName": "id"
    }
  ]
}
```

The user is shown:

> You do not have permission to perform this action. Contact an administrator if you believe this is a mistake.

They had the permission. What they needed to know was to remove the members first — which the response says, in a field the interceptor never reaches. Instead they go and ask an administrator for a right they already hold, and the administrator finds nothing wrong.

This is not specific to groups. Every state-machine rule in Fineract comes back this way: closing a client with open loans, disbursing before approval, undoing a transaction that has been reversed. The whole class currently reports as a permissions problem.

## Where it is

`src/app/core/interceptors/error.interceptor.ts`, in `messageFor`. The 403 branch returns before the `errors[]` handling that would have produced the right message:

```ts
// Fineract's 403 body describes the missing permission in backend terms ("NOT_ALLOWED",
// a permission code) which means nothing to the user. ...
if (error.status === 403) {
  return i18n.translate('COMMON.ERRORS.FORBIDDEN');
}

if (error.error?.errors && Array.isArray(error.error.errors) && error.error.errors.length > 0) {
  // ← this is what should run for a domain-rule 403, and never does
```

The comment is right about *authorization* 403s — that body really is unhelpful, and the friendly message is the correct treatment. The bug is that it also swallows the domain-rule 403, which is the informative one.

## Reproducing it

Against a local Fineract:

1. Create a client and activate it.
2. Create a group, activate it, and add the client as a member.
3. Add a value to the `GroupClosureReason` code (System → Manage Codes) — it ships empty.
4. Open the group and choose Close.

Expected: the platform's reason. Actual: the permissions message.

`e2e/group-membership.spec.ts` drives exactly this sequence and currently asserts only that the command is refused and the group is left Active, with a comment pointing here — it should assert the message once this is fixed.

## Describing the change

Distinguish the two. A domain-rule violation is identifiable without guessing:

- `userMessageGlobalisationCode` is `validation.msg.domain.rule.violation`, and
- `errors[]` is populated with entries carrying a real `defaultUserMessage`.

An authorization 403 has neither. So the conservative fix is to let a 403 fall through to the existing `errors[]` handling when that array has content, and keep `COMMON.ERRORS.FORBIDDEN` otherwise:

```ts
const isDomainRuleViolation =
  error.error?.userMessageGlobalisationCode === 'validation.msg.domain.rule.violation' ||
  (Array.isArray(error.error?.errors) && error.error.errors.length > 0);

if (error.status === 403 && !isDomainRuleViolation) {
  return i18n.translate('COMMON.ERRORS.FORBIDDEN');
}
```

Deliberately conservative: a genuine permission 403 that happens to carry an `errors[]` array still shows the friendly message only if the array is empty, and anything unrecognised keeps today's behaviour. Nothing gets *worse* than it is now.

`error.interceptor.spec.ts` already covers the 403 path and the `errors[]` stacking separately — this needs a case for the two combined. Please add one asserting the group-close body above produces the platform's message rather than the permissions text.

## Scope

In scope: the 403 branch in `messageFor`, and a unit spec for it.

Out of scope: the 401 session handling beside it, changing `COMMON.ERRORS.FORBIDDEN` itself, and any per-screen error handling — this is one interceptor and every screen benefits.

## Getting started

- `src/app/core/interceptors/error.interceptor.ts` and its spec.
- `npm test` must pass; `npm run lint` and `npm run build` too.
- Worth reading the existing spec first — the 403 case there encodes the reasoning this change has to preserve.

Guflly:
## What and why

Show Fineract's domain-rule explanation for 403 responses while keeping the generic permission message for authorization failures.

Closes #258

## Verification

- Focused interceptor suite: 13 passed
- npm run lint
- npm run build
- npm run check:icons
- npm run i18n:check

## Screenshots

Not applicable.

## Checklist

- [x] No new component or service code was added.
- [x] No user-facing strings were added.
- [x] I added regression coverage for the domain-rule response.
- [x] No UI workflow changed.

### DIFF
--- a/src/app/core/interceptors/error.interceptor.spec.ts
+++ b/src/app/core/interceptors/error.interceptor.spec.ts
@@ -268,6 +268,33 @@ describe('errorInterceptor', () => {
       expect(routerSpy.navigate).not.toHaveBeenCalled();
     });
 
+    it('should report the platform reason for a domain rule violation', () => {
+      httpClient.get(testUrl, authorized).subscribe({ error: () => expect().nothing() });
+
+      httpTestingController.expectOne(testUrl).flush(
+        {
+          userMessageGlobalisationCode: 'validation.msg.domain.rule.violation',
+          defaultUserMessage: 'Errors contain reason for domain rule violation.',
+          errors: [
+            {
+              defaultUserMessage:
+                'Group cannot be closed because of active clients associated with it.',
+              userMessageGlobalisationCode: 'error.msg.Group.close.active.clients.exist',
+              parameterName: 'id',
+            },
+          ],
+        },
+        { status: 403, statusText: 'Forbidden' },
+      );
+
+      expect(notificationsSpy.error).toHaveBeenCalledWith(
+        'Errors contain reason for domain rule violation.\n\n' +
+          '\u2022 [id] Group cannot be closed because of active clients associated with it.',
+      );
+      expect(authSpy.logout).not.toHaveBeenCalled();
+      expect(routerSpy.navigate).not.toHaveBeenCalled();
+    });
+
     it('should stay silent when the caller opts out', () => {
       httpClient
         .get(testUrl, { context: skipErrorToast() })
--- a/src/app/core/interceptors/error.interceptor.ts
+++ b/src/app/core/interceptors/error.interceptor.ts
@@ -55,14 +55,19 @@ function messageFor(error: HttpErrorResponse, i18n: I18nAdapter): string {
     return `Error: ${error.error.message}`;
   }
 
+  const hasValidationErrors = Array.isArray(error.error?.errors) && error.error.errors.length > 0;
+  const isDomainRuleViolation =
+    error.error?.userMessageGlobalisationCode === 'validation.msg.domain.rule.violation' ||
+    hasValidationErrors;
+
   // Fineract's 403 body describes the missing permission in backend terms ("NOT_ALLOWED",
   // a permission code) which means nothing to the user. What they can act on is that this
   // account lacks the right, not which internal code was checked.
-  if (error.status === 403) {
+  if (error.status === 403 && !isDomainRuleViolation) {
     return i18n.translate('COMMON.ERRORS.FORBIDDEN');
   }
 
-  if (error.error?.errors && Array.isArray(error.error.errors) && error.error.errors.length > 0) {
+  if (hasValidationErrors) {
     const stacked = error.error.errors
       .map((err: Record<string, unknown>) => {
         const msg = err['developerMessage'] || err['defaultUserMessage'] || 'Validation error';


Aman-Mittal:
@Guflly please sign your commits

Aman-Mittal:
Thanks for picking this up — the diagnosis is right, and letting a domain-rule 403 through to the stacked-message branch is the correct shape for the fix. One thing to change before it lands, and it is only visible against a real platform.

## The stacked branch prefers `developerMessage`

The branch this now routes 403s into already reads:

```ts
const msg = err['developerMessage'] || err['defaultUserMessage'] || 'Validation error';
```

`developerMessage` first. Reproduced against a running Fineract just now:

```
DELETE /codes/15/codevalues/24   →   403
  errors[0].developerMessage   : "The request caused a data integrity issue to be fired by the database."
  errors[0].defaultUserMessage : "This code value is in use"
```

So after this change that user stops seeing *"You do not have permission to perform this action"* and starts seeing *"The request caused a data integrity issue to be fired by the database."* — which is not an improvement, and is the same category of problem the comment left in place is warning about. The identical preference sits in the fallback a few lines below.

Flipping it to `defaultUserMessage || developerMessage` in both places gives *"This code value is in use"*, which is the message the platform wrote for the user.

The new test does not catch this because its fixture has no `developerMessage` on the error entry. Real Fineract bodies always carry one — the group-closure example in the test looks like this in practice:

```json
{
  "errors": [
    {
      "developerMessage": "A loan with money outstanding cannot be closed",
      "defaultUserMessage": "A loan with money outstanding cannot be closed",
      "userMessageGlobalisationCode": "error.msg.loan.close.loan.has.money.outstanding",
      "parameterName": "id"
    }
  ]
}
```

A fixture where the two strings **differ** would pin the behaviour that matters.

## Two smaller notes

**`isDomainRuleViolation` collapses to `hasValidationErrors`.** Since it is `code === '…' || hasValidationErrors`, and the `FORBIDDEN` branch is skipped whenever either holds, the globalisation-code half never changes an outcome. Worth dropping it, or keeping it and dropping the `||` — as written it reads like it does more than it does.

There is also a gap between the two: when the code matches but `errors[]` is empty, this skips `FORBIDDEN`, skips the stacked branch, and lands on the top-level `developerMessage` — *"Request was understood but caused a domain rule violation."*

**A question rather than a claim:** does the platform's genuine permission-denied 403 also carry an `errors[]` array? If it does, this change routes it into the stacked branch too, and the user would be shown the `NOT_ALLOWED` permission code that the original guard existed to hide. I could not construct that case here without a restricted user, so I have not asserted it either way — but it seems worth one test before merge, since it is the only path that could make things worse rather than better.

Happy to be wrong on any of this; the underlying change is one the app needs.

Aman-Mittal:
closing this due to inactivity pls reopen when you are back

PR #270, opened 2026-08-08
fix(routing): redirect bare calendar and meeting paths

### LINKED ISSUE #264: /calendars and /meetings dead-end on a blank page instead of redirecting like every other unknown URL
## Summary

Opening `/calendars` or `/meetings` leaves you on a page with the header and sidebar drawn and **nothing at all** in the content area. No table, no message, no error. The URL stays put, so it does not look like a failed navigation — it looks like a screen that failed to load.

Every other unknown URL in the app is handled: `path: '**'` in `src/app/app.routes.ts` redirects to the dashboard.

Measured on a running app, content area character count in brackets:

| URL | Lands on | Content |
|---|---|---|
| `/zzz-does-not-exist` | `/dashboard` | 482 chars |
| `/clients/zzz-nope` | `/dashboard` | 482 chars |
| `/calendars` | `/calendars` | **0 chars** |
| `/meetings` | `/meetings` | **0 chars** |

## Cause

`src/app/features/calendars/calendars.routes.ts` and `src/app/features/meetings/meetings.routes.ts` define only entity-scoped routes:

```ts
export const CALENDARS_ROUTES: Routes = [
  { path: ':entityType/:entityId', ... },
  { path: ':entityType/:entityId/create', ... },
  { path: ':entityType/:entityId/edit/:id', ... },
];
```

There is no `''` route. The parent route in `app.routes.ts` still matches `/calendars` and lazy-loads this list, so the router considers the URL handled and never reaches the app-level `'**'`. It then has no child to render, and draws nothing.

## Fix

Give each file an index route so the URL resolves the same way every other unmatched URL does:

```ts
export const CALENDARS_ROUTES: Routes = [
  {
    path: '',
    pathMatch: 'full',
    redirectTo: '/dashboard',
  },
  { path: ':entityType/:entityId', ... },
  ...
];
```

**Verified**: applied to `calendars.routes.ts` only, with `meetings.routes.ts` left untouched as a control. `/calendars` then redirects to the dashboard on load, and `/meetings` still shows the blank page. Apply the same change to both files.

Note when testing: navigating there from inside the app can reuse the router config already in memory, so load the URL fresh (or hard-reload) rather than clicking through.

## Context, not part of this issue

While confirming this I found that **nothing in the app links to these routes at all** — no sidebar entry, no `routerLink` anywhere. The four components behind them (`calendars-list`, `calendar-form`, `meetings-list`, `meeting-form`) currently have no way in. Giving them an entry point is a separate, larger question that overlaps the detail-view work in #180 and #181, so please do not take it on here. This issue is only about the dead end.

## Business Value

A blank screen is the worst thing a URL can do, because it gives the user nothing to act on — they cannot tell whether the page is broken, whether they lack permission, or whether they mistyped. Anyone landing here from a stale bookmark, a shared link, or a typed URL is simply stuck, and the natural next step is to report the app as broken. Redirecting costs nothing and puts them somewhere they can carry on from.

It also closes an inconsistency: the app already decided what an unknown URL should do, and these two are the only paths that quietly opt out of it.

## How to check your work

1. `npm start`, sign in, then load `https://localhost:4200/calendars` and `https://localhost:4200/meetings` directly in the address bar. Both should land on the dashboard.
2. `/calendars/groups/1` and `/meetings/groups/1` should still open their screens.
3. `npm test` and `npm run lint` still pass.

## Good first issue

Two files, four lines each, no logic and no API involved, and the result is visible in the address bar.

alloutflo:
## What and why

Redirect the bare `/calendars` and `/meetings` feature roots to `/dashboard`, matching the application's existing fallback behavior. Those lazy-loaded route trees previously had no empty-path child, so both URLs matched the feature prefix but left the outlet blank instead of reaching the global wildcard redirect.

The existing entity-scoped list, create, and edit routes are unchanged.

Closes #264

## Verification

- Focused RouterTestingHarness specs: 4 passed
- CI full Karma suite: 821 passed
- Exercised both redirects with Angular Router 22.0.7 against the real feature route arrays and pinned representative entity-scoped routes against regressions
- `npm run lint`
- HTML ESLint
- `npm run format:check`
- `npm run i18n:check`
- `npm run check:icons`
- `npm run build`
- API client regeneration and drift check
- ASF license-header check
- Production dependency license check
- `npm audit --audit-level=high --omit=dev` (0 vulnerabilities)

`npm run api:surface` still reports the same three pre-existing manifest gaps reproduced on clean `main` at `4c884831`; this change does not touch the generated API or its surface manifest.

## Screenshots

Not applicable. This is a route-configuration fix with no visual change.

## Checklist

- [x] Not applicable: no component or service code was added.
- [x] Not applicable: no user-facing strings were added.
- [x] I added regression coverage for both feature roots.
- [x] Not applicable: no Fineract backend interaction changed; Angular Router recognition was verified directly.

### DIFF
--- a/src/app/features/calendars/calendars.routes.spec.ts
+++ b/src/app/features/calendars/calendars.routes.spec.ts
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Component } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { provideRouter, Router } from '@angular/router';
+import { RouterTestingHarness } from '@angular/router/testing';
+import { CALENDARS_ROUTES } from './calendars.routes';
+
+@Component({
+  standalone: true,
+  template: '',
+})
+class RouteStub {}
+
+describe('CALENDARS_ROUTES', () => {
+  beforeEach(() => {
+    const routes = CALENDARS_ROUTES.map((route) =>
+      route.path === ':entityType/:entityId' ? { path: route.path, component: RouteStub } : route,
+    );
+
+    TestBed.configureTestingModule({
+      providers: [
+        provideRouter([
+          { path: 'calendars', children: routes },
+          { path: 'dashboard', component: RouteStub },
+        ]),
+      ],
+    });
+  });
+
+  it('redirects the feature root to the dashboard', async () => {
+    const harness = await RouterTestingHarness.create();
+
+    await harness.navigateByUrl('/calendars');
+
+    expect(TestBed.inject(Router).url).toBe('/dashboard');
+  });
+
+  it('keeps entity-scoped routes reachable', async () => {
+    const harness = await RouterTestingHarness.create();
+
+    await harness.navigateByUrl('/calendars/groups/1');
+
+    expect(TestBed.inject(Router).url).toBe('/calendars/groups/1');
+  });
+});
--- a/src/app/features/calendars/calendars.routes.ts
+++ b/src/app/features/calendars/calendars.routes.ts
@@ -21,6 +21,11 @@ import { Routes } from '@angular/router';
 
 export const CALENDARS_ROUTES: Routes = [
   {
+    path: '',
+    pathMatch: 'full',
+    redirectTo: '/dashboard',
+  },
+  {
     path: ':entityType/:entityId',
     loadComponent: () => import('./calendars-list.component').then((m) => m.CalendarsListComponent),
   },
--- a/src/app/features/meetings/meetings.routes.spec.ts
+++ b/src/app/features/meetings/meetings.routes.spec.ts
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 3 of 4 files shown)


Aman-Mittal:
The change is right, and I checked it rather than eyeballing it. Applied this exact diff locally against a running app:

| URL | Result |
|---|---|
| `/calendars` | → `/dashboard` |
| `/meetings` | → `/dashboard` |
| `/calendars/groups/1` | stays put, `app-calendars-list` mounts |
| `/meetings/groups/1` | stays put |

The last two are the ones worth knowing about, because they are the regression this could plausibly have caused and nothing in the PR covers them. `pathMatch: 'full'` is doing its job — thank you for including it, an unqualified `redirectTo` here would have swallowed the entity-scoped routes and that failure would not have been obvious.

## One thought on the specs

They currently assert that the route config object equals the literal written in the source file:

```ts
const rootRoute = CALENDARS_ROUTES.find(({ path }) => path === '');
expect(rootRoute).toEqual({ path: '', pathMatch: 'full', redirectTo: '/dashboard' });
```

That guards against someone deleting the route by accident, which is worth something. But it cannot fail for any reason other than the literal being edited — it would pass just as happily if the redirect never worked at runtime, so it does not really pin the behaviour this PR is about.

`RouterTestingHarness` can assert the actual navigation, and it is not much longer:

```ts
@Component({ standalone: true, template: '' })
class DashboardStub {}

it('sends the bare feature path to the dashboard', async () => {
  TestBed.configureTestingModule({
    providers: [
      provideRouter([
        { path: 'calendars', children: CALENDARS_ROUTES },
        { path: 'dashboard', component: DashboardStub },
      ]),
    ],
  });

  const harness = await RouterTestingHarness.create();
  await harness.navigateByUrl('/calendars');

  expect(TestBed.inject(Router).url).toBe('/dashboard');
});
```

Note the stub dashboard route — `redirectTo: '/dashboard'` is absolute, so it resolves against the whole router config rather than this feature's, and a test wiring up only `CALENDARS_ROUTES` would have nothing to land on.

The same shape also lets you assert the case that actually worries me, in one more line: navigating to `/calendars/groups/1` and expecting the URL to stay put.

Entirely your call — this is a suggestion, not a blocker. The change itself is correct and I would be happy to see it merged as is.

Aman-Mittal:
@alloutflo Please sign your commits

Aman-Mittal:
@alloutflo 

This user has not yet uploaded their public signing key.
[Learn about vigilant mode](https://docs.github.com/github/authenticating-to-github/displaying-verification-statuses-for-all-of-your-commits)

alloutflo:
Thanks — I’ve now uploaded the public SSH signing key, and the current commit `e8be03f` is shown as Verified. I also incorporated your RouterTestingHarness suggestion, including regression coverage for the entity-scoped routes. All checks are green.

Aman-Mittal (approved):
LGTM thanks for contribution

PR #275, opened 2026-08-08
ci(e2e): shard mocked Playwright tests

### LINKED ISSUE #259: Shard the Playwright e2e suite across a job matrix, as apache/fineract does
## Business value

The e2e suite runs on a single worker and is now the longest thing on the critical path of every pull request.

Measured on this repository:

| Project | Tests | Files | CI wall-clock |
| --- | --- | --- | --- |
| `mocked` | 220 | 13 | ~11m45s (job `92945974006`, 17:31:53 → 17:43:38) |
| `backend` | 23 | 10 | longer — brings up a Fineract stack first |

`playwright.config.ts` pins `workers: process.env.CI ? 1 : undefined`, deliberately: the specs create real records against a shared Fineract, and parallel workers were producing interference. So the suite cannot be sped up by adding workers inside one job — the parallelism has to come from running *disjoint sets of tests in separate jobs*, each with its own backend.

That is exactly what the platform repository does. `apache/fineract`'s [`build-e2e-tests.yml`](https://github.com/apache/fineract/blob/develop/.github/workflows/build-e2e-tests.yml) runs a **20-way shard matrix**:

```yaml
strategy:
  fail-fast: false
  matrix:
    shard_index: [1, 2, …, 20]
    total_shards: [20]
```

with `scripts/split-features.sh` dividing the feature files, a **fresh docker stack per shard**, and per-shard artifacts uploaded on failure. The shape is proven in the same project family, on the same kind of test.

Why this matters beyond wall-clock: a suite slow enough to be annoying is a suite people stop waiting for. The cost is paid on every push, and this repo is under active concurrent development.

## What makes it tractable here

Playwright has native sharding — no equivalent of `split-features.sh` is needed:

```
npx playwright test --project=mocked --shard=${{ matrix.shard }}/${{ matrix.total }}
```

It splits by test and balances across shards. Combined with the `blob` reporter, the shards' results merge back into a single report:

```
# per shard
npx playwright test --shard=1/4 --reporter=blob
# after the matrix
npx playwright merge-reports --reporter=html ./all-blob-reports
```

## Describing the change

**Start with the `mocked` project only.** It needs no Fineract and no seeding, so a shard is just another checkout — the sharding is nearly free and the payoff is immediate. Four shards should take ~11m45s to roughly 3–4m including setup.

- Add `strategy.matrix` with `fail-fast: false` to the `mocked` job in `.github/workflows/e2e.yml`.
- Pass `--shard=${{ matrix.shard }}/${{ matrix.total }}`.
- Switch the shard reporter to `blob`, upload each shard's blob as an artifact, and add a dependent job that downloads them and runs `merge-reports`.
- Keep the existing PR summary comment working off the merged report rather than off one shard, or it will report a quarter of the results as if they were all of them.

**Treat the `backend` project as a separate decision, in a follow-up.** It is 23 tests, and each shard would pay the full docker bring-up — database, `init-db.sql`, waiting on the actuator health endpoint — so naive sharding could make it *slower*, not faster. It also needs care that mocked does not:

- The specs create real records and some depend on state they set up earlier in the same file. Playwright's sharding splits by test, so two tests in one file can land in different shards. `fullyParallel: true` is already set, which means the suite is *supposed* to tolerate that, but it has never actually been run that way — this needs verifying rather than assuming.
- Shards must not share one Fineract. `apache/fineract` gives each shard its own stack for exactly this reason. Sharing one would reintroduce the interference `workers: 1` exists to prevent.
- One known example to check: `teller-cash-management.spec.ts` allocates a cashier and the platform refuses overlapping date ranges for the same staff member across *any* teller, which is why that spec creates fresh staff per run. Similar cross-test coupling elsewhere would surface only under sharding.

So: shard `mocked` first, measure, then decide on `backend` with numbers in hand.

## Scope

In scope: sharding the `mocked` e2e job, blob reporting, report merging, and keeping the PR summary accurate.

Out of scope for the first PR: sharding the `backend` job, changing `workers` or `retries`, and any change to the specs themselves.

## Notes for whoever picks this up

- `.github/workflows/e2e.yml` holds both jobs; the concurrency group at the top already cancels superseded runs and should be left alone.
- `playwright.config.ts` sets `reporter: 'html'` — the shard jobs need `blob`, which is best passed on the command line rather than changed globally, so a local run still produces the HTML report developers expect.
- `outputDir` is deliberately outside the repo (the dev server's file watcher races Playwright's artifact deletion and dies); keep that.
- Verify by comparing the merged report's total against `npx playwright test --list --project=mocked` — currently 220. A sharding mistake shows up as a *smaller* total that still reports green, which is the failure mode to watch for.

Shreyas0786:
## What and why

Shards the mocked Playwright project across four GitHub Actions matrix jobs, then merges their blob reports in a dependent job. This reduces the mocked-suite wall-clock time while preserving the combined HTML/JSON report and PR summary. The backend project remains unchanged.

Closes #259

## Verification

- `npx playwright test --list --project=mocked` — 220 tests in 13 files
- Verified four shards contain 55 tests each, totaling 220
- Dry-ran four list-only blob reports and merged them successfully — merged JSON contained 220 tests across 13 suites
- `npm run format:check`
- `npm run lint`
- `npm run build`
- `bash scripts/check-license.sh`
- No UI or real Fineract backend was exercised; this is a workflow-only change

## Screenshots

Not applicable — workflow-only change with no visual UI changes.

## Checklist

- [x] New component or service code uses the adapter boundary in `src/app/core/adapters/` instead of direct browser globals or imperative third-party APIs. Not applicable; no component or service code changed.
- [x] User-facing strings use translation keys. Not applicable; no user-facing strings changed.
- [x] I added or updated tests appropriate to this change, or explained why tests were not needed. The existing 220-test suite was verified across four shards and through a merged-report dry run.
- [x] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. Not applicable; no UI workflow changed, and the backend job remains unchanged.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -136,7 +136,7 @@ jobs:
         env:
           PLAYWRIGHT_HTML_OPEN: 'never'
           PLAYWRIGHT_JSON_OUTPUT_NAME: 'playwright-results.json'
-        run: npx playwright merge-reports --reporter=html,json ./all-blob-reports
+        run: npx playwright merge-reports --reporter=line,html,json ./all-blob-reports
 
       - name: Upload merged report and videos
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -36,21 +36,25 @@ permissions: {}
 # The suite is split in two so a broken mock does not wait behind a Fineract
 # boot, and so the slow half is not on the critical path for the fast one.
 #
-#   mocked  — 188 tests that stub the backend with page.route(). Needs no
+#   mocked  — 220 tests that stub the backend with page.route(). Needs no
 #             Fineract at all, so it skips the docker bring-up entirely and
 #             reports in a couple of minutes.
-#   backend — 16 tests that drive a real Fineract end to end. Slow, and the
+#   backend — 23 tests that drive a real Fineract end to end. Slow, and the
 #             only half that needs the stack.
 #
 # The split lives in playwright.config.ts as two projects, so `--project=mocked`
 # and `--project=backend` mean the same thing locally as they do here.
 jobs:
   mocked:
-    name: E2E (mocked backend)
+    name: E2E (mocked backend, shard ${{ matrix.shard }}/${{ matrix.total }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4]
+        total: [4]
     permissions:
       contents: read # checkout
-      pull-requests: write # upsert this job's summary comment
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4
         with:
@@ -81,19 +85,63 @@ jobs:
 
       - name: Run mocked E2E tests
         env:
+          SHARD_INDEX: ${{ matrix.shard }}
+          SHARD_TOTAL: ${{ matrix.total }}
+          PLAYWRIGHT_OUTPUT_DIR: ${{ runner.temp }}/e2e-output
+        run: npx playwright test --project=mocked --shard="$SHARD_INDEX/$SHARD_TOTAL" --reporter=blob
+
+      - name: Upload blob report
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        if: ${{ !cancelled() }}
+        with:
+          name: playwright-blob-mocked-${{ matrix.shard }}
+          path: blob-report/
+          if-no-files-found: warn
+          retention-days: 1
+
+  mocked-report:
+    name: E2E (mocked backend report)
+    if: ${{ !cancelled() }}
+    needs: mocked
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # checkout
+      pull-requests: write # upsert the merged summary comment
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v4
+        with:
+          persist-credentials: false
+
+      - name: Set up Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 'lts/*'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Download blob reports
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          path: all-blob-reports
+          pattern: playwright-blob-mocked-*
+          merge-multiple: true
+
+      - name: Merge mocked E2E reports
+        env:
           PLAYWRIGHT_HTML_OPEN: 'never'
           PLAYWRIGHT_JSON_OUTPUT_NAME: 'playwright-results.json'
-          PLAYWRIGHT_OUTPUT_DIR: ${{ runner.temp }}/e2e-output
-        run: npx playwright test --project=mocked --reporter=line,html,json
+        run: npx playwright merge-reports --reporter=html,json ./all-blob-reports
 
-      - name: Upload report and videos
+      - name: Upload merged report
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-        if: always()
+        if: ${{ !cancelled() }}
         with:
           name: playwright-report-mocked
           path: |
             playwright-report/
-            ${{ runner.temp }}/e2e-output/
+            playwright-results.json
           if-no-files-found: warn
           retention-days: 7
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -36,14 +36,18 @@ permissions: {}
 # The suite is split in two so a broken mock does not wait behind a Fineract
 # boot, and so the slow half is not on the critical path for the fast one.
 #
-#   mocked  — 220 tests that stub the backend with page.route(). Needs no
+#   mocked  — tests that stub the backend with page.route(). Needs no
 #             Fineract at all, so it skips the docker bring-up entirely and
 #             reports in a couple of minutes.
-#   backend — 23 tests that drive a real Fineract end to end. Slow, and the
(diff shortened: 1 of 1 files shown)


Aman-Mittal (commented):
Nice piece of work — the blob-report/merge shape is exactly the right way to shard Playwright, the `pull-requests: write` permission correctly moved off the four shard jobs onto the single job that actually posts a comment, and `fail-fast: false` means one red shard doesn't hide the other three. All 22 checks are green here, and the four shards did run.

Two things about **how the result is reported** that I'd want settled before this merges. Neither is a defect in the YAML — both are about what happens to the repo once the job names change.

---

### 1. The required status check `E2E (mocked backend)` no longer exists

The job is now named `E2E (mocked backend, shard ${{ matrix.shard }}/${{ matrix.total }})`, so it reports as four checks:

```
E2E (mocked backend, shard 1/4)   ← new
E2E (mocked backend, shard 2/4)
E2E (mocked backend, shard 3/4)
E2E (mocked backend, shard 4/4)
E2E (mocked backend)              ← gone
```

If `E2E (mocked backend)` is listed as a required status check on `main`, every PR after this one waits forever on a check that will never be reported. I can't read the protection settings to confirm (the endpoint is admin-only), so this needs a maintainer to check the branch protection rules and swap the old name for the four new ones **in the same change window as the merge**.

### 2. `E2E (mocked backend report)` is green even when tests fail — so it can't be the required check

This is the one worth being explicit about, because the obvious fix for (1) — "just require the single report job instead of four shards" — silently disables the E2E gate.

I checked rather than assumed. A blob report containing one passing and one failing test, merged:

```
$ npx playwright test --reporter=blob     # 1 failed, 1 passed
  exit = 1
$ npx playwright merge-reports --reporter=line ./all-blob
  exit = 0        ← the failure is in the report, and the job still passes
```

`merge-reports` reports; it doesn't adjudicate. Combined with `if: ${{ !cancelled() }}`, `mocked-report` succeeds whenever the merge itself succeeds, no matter what the shards found. That's correct behaviour for this job — you want the report even on failure — but it means **the four shard jobs have to stay the required checks**, and `mocked-report` must not be substituted for them.

Would you mind adding a line to the comment block above `jobs:` saying so? The workflow already explains why the suite is split in two, and this is the same kind of thing a future reader would otherwise have to rediscover the hard way.

---

### Smaller notes, none blocking

- **`PLAYWRIGHT_OUTPUT_DIR` in the shard step is now redundant.** `playwright.config.ts` already defaults `outputDir` to `os.tmpdir()`, and nothing uploads that path any more. Harmless either way, so keep it if you'd rather be explicit.

- **The raw `e2e-output/` upload is gone**, which I think is fine but worth confirming you intended: traces, videos and screenshots are attached into the blob and come back out in the merged HTML report, so they're still reachable — just via the report rather than as loose files in the artifact.

- **The tests-per-shard split should stay balanced.** `fullyParallel: true` means Playwright shards by individual test rather than by file, so the 55/55/55/55 you measured isn't a coincidence of the current file sizes and won't drift as files grow unevenly.

- **The header comment counts are now stale in the other direction.** You've updated them to 220/23, and `#276` is about to add 15 mocked tests. Not your problem to fix — just flagging that these numbers are load-bearing documentation that nothing verifies, which is arguably its own small issue.

Happy to re-review once the branch-protection question has an owner.

.github/workflows/e2e.yml
```diff
@@ -80,20 +84,64 @@ jobs:
...
+      - name: Merge mocked E2E reports
         env:
           PLAYWRIGHT_HTML_OPEN: 'never'
           PLAYWRIGHT_JSON_OUTPUT_NAME: 'playwright-results.json'
-          PLAYWRIGHT_OUTPUT_DIR: ${{ runner.temp }}/e2e-output
-        run: npx playwright test --project=mocked --reporter=line,html,json
+        run: npx playwright merge-reports --reporter=html,json ./all-blob-reports
 
-      - name: Upload report and videos
+      - name: Upload merged report
```
Aman-Mittal: merged report and videos*

Aman-Mittal: Same as above as it also uploads video of passed tests.

.github/workflows/e2e.yml
```diff
@@ -36,21 +36,25 @@ permissions: {}
 # The suite is split in two so a broken mock does not wait behind a Fineract
 # boot, and so the slow half is not on the critical path for the fast one.
 #
-#   mocked  — 188 tests that stub the backend with page.route(). Needs no
+#   mocked  — 220 tests that stub the backend with page.route(). Needs no
```
Aman-Mittal: We can remove this entirely as new tests are being added for each feature

Aman-Mittal:
@Shreyas0786 Also rebase against main

.github/workflows/e2e.yml
Shreyas0786: Updated the step name to Upload merged report and videos.

.github/workflows/e2e.yml
Shreyas0786: Removed the fixed test counts from both project descriptions.

Shreyas0786:
Rebased onto the latest main and addressed the review comments in 95d1862. I also documented that branch protection must require all four shard checks rather than the report job. @Aman-Mittal, could you confirm the old E2E (mocked backend) requirement will be replaced with the four shard checks when this merges?

Aman-Mittal:
> Rebased onto the latest main and addressed the review comments in [95d1862](https://github.com/apache/fineract-backoffice-ui/commit/95d18624ddd043f936d24705022d0251d81b42b1). I also documented that branch protection must require all four shard checks rather than the report job. @Aman-Mittal, could you confirm the old E2E (mocked backend) requirement will be replaced with the four shard checks when this merges?

Yes

.github/workflows/e2e.yml
```diff
@@ -80,20 +88,64 @@ jobs:
...
+          pattern: playwright-blob-mocked-*
+          merge-multiple: true
+
+      - name: Merge mocked E2E reports
         env:
           PLAYWRIGHT_HTML_OPEN: 'never'
           PLAYWRIGHT_JSON_OUTPUT_NAME: 'playwright-results.json'
-          PLAYWRIGHT_OUTPUT_DIR: ${{ runner.temp }}/e2e-output
-        run: npx playwright test --project=mocked --reporter=line,html,json
+        run: npx playwright merge-reports --reporter=html,json ./all-blob-reports
```
Aman-Mittal: --reporter=line,html,json ac it will show only
.........

i will not able to confirm what test ran on that shard

Shreyas0786: Updated the merged reporter list to line,html,json, so the job log shows the tests executed across the shards.

Aman-Mittal (approved):

PR #277, opened 2026-08-08
docs: document the commit-signing requirement for contributors

### LINKED ISSUE #271: Commits must be signed to be merged, and nothing tells a contributor that until their PR is blocked
## Summary

Branch protection on `main` requires signed commits, so an unsigned one cannot be merged. That rule is enforced but never stated: a contributor finds out at the very end, once the work is written, reviewed and pushed, and the merge is blocked.

Searched the whole contributor-facing surface — `README.md`, `CONTRIBUTING.md`, `.github/pull_request_template.md`, `.github/ISSUE_TEMPLATE/`, `DOCS/` — and **signing is not mentioned anywhere**. There is no `gpgsign`, no `Signed-off-by`, no mention of GPG or SSH signing. `CONTRIBUTING.md` has a "Pull Request Guidelines" section, which is exactly where someone would look, and it says nothing about it.

So the requirement is discoverable only by hitting it.

## It is happening now, not hypothetically

Both community pull requests currently open carry unsigned commits:

| PR | Commit | Verification |
|---|---|---|
| #270 `fix(routing): redirect bare calendar and meeting paths` | `666a3a18` | **unsigned** |
| #260 `fix: show domain rule errors for 403 responses` | `7a07f616` | **unsigned** |

Both are small, correct, well-made contributions. Both authors did everything asked of them by the documentation, because the documentation does not ask for this. Whatever happens to those two PRs, the next contributor will hit the same wall for the same reason.

## What to change

1. **A short section in `CONTRIBUTING.md`**, under or beside "Pull Request Guidelines", covering:
   - that `main` requires signed commits and an unsigned one cannot be merged
   - the one-time setup for **both** GPG and SSH signing — SSH is considerably less work and many contributors already have a key
   - `git config commit.gpgsign true` so it is not a per-commit flag to remember
   - how to fix a branch that is already unsigned: `git rebase --exec 'git commit --amend --no-edit -S' <base>`, since arriving here after the fact is the common case
2. **A line in `.github/pull_request_template.md`**, so it is on screen while the PR is being opened rather than only in a file people read once.
3. Optionally, a pointer from the README's contributing link.

Links worth including: [GitHub's signing setup docs](https://docs.github.com/en/authentication/managing-commit-signature-verification) and the SSH signing section in particular.

## Business Value

This is the first impression the project makes on someone who has already decided to help. They write the fix, open the PR, get through review — and only then discover a requirement nobody mentioned, with the remedy being to rewrite their commits. That is the point where a first-time contributor quietly gives up, and the project never learns why.

It also costs maintainer time on every single new contributor: the same explanation, written by hand, after review, forever. Writing it down once converts a recurring personal request into something the contributor handles before anyone looks at the branch.

And it removes an unfair asymmetry. Regular contributors have signing configured and never think about it; the rule is invisible to precisely the people it does not affect, and a wall for the people it does.

## Not in scope

Adding a CI check. Branch protection already enforces this — a check would only repeat the same refusal slightly earlier, and the gap here is that nobody is told, not that nobody is stopped. If a check is wanted later so the failure is visible *before* review rather than at merge, that is worth its own issue and its own argument.

## Good first issue

Documentation only, no code, and the change can be checked by reading it. One caveat worth naming honestly: whoever takes this will need to sign their own commits to get it merged — which makes them exactly the right person to write the instructions, since they will have just followed them.

AshharAhmadKhan:
## What and why

The commit-signing requirement enforced on main is not documented anywhere a contributor would see it before submitting a pull request. CONTRIBUTING.md's Pull Request Guidelines section does not mention it, and neither does the pull request template. As a result, a contributor can follow the existing documentation exactly and still have their pull request blocked at merge time, with no prior indication that signing was required.

This adds a Commit Signing section to CONTRIBUTING.md covering the requirement and the one-time setup, a bullet in Pull Request Guidelines pointing to it, and a comment in the pull request template so the requirement is visible before work begins rather than discovered at merge.

Closes #271

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -17,6 +17,8 @@ specific language governing permissions and limitations
 under the License.
 -->
 
+<!-- Commits must be signed to merge — see CONTRIBUTING.md#commit-signing if you haven't set this up. -->
+
 ## What and why
 
 <!-- One or two sentences explaining what changed and why. -->
@@ -42,3 +44,4 @@ Closes #
 - [ ] User-facing strings use translation keys.
 - [ ] I added or updated tests appropriate to this change, or explained why tests were not needed.
 - [ ] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant.
+- [ ] Commits are signed — see [Commit Signing](CONTRIBUTING.md#commit-signing) in CONTRIBUTING.md.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,4 +119,26 @@ relying on transitive resolution, so the audit sees them.
   back-office UI are tracked here, not in Jira — the ASF Jira project is for
   [apache/fineract](https://github.com/apache/fineract), the platform itself.
 - Ensure CI checks pass.
+- Commits must be signed (GPG) — see [Commit Signing](#commit-signing) below.
 - New features should include unit tests.
+
+## Commit Signing
+
+`main` requires signed commits — an unsigned commit cannot be merged. Set this up
+before you start work, not after.
+
+```bash
+git config user.signingkey <your-gpg-key-id>
+git config commit.gpgsign true
+```
+
+`commit.gpgsign true` makes signing automatic for every commit, so it isn't a flag
+you have to remember. See GitHub's docs for full setup instructions:
+[commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification).
+
+**If you already have unsigned commits on your branch**, sign them retroactively
+instead of starting over:
+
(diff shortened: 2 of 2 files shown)


AshharAhmadKhan:
Hey @Aman-Mittal , please review this when you get a chance, thanks!

.github/pull_request_template.md
Aman-Mittal, line 46: Also add it to checklist to with docs link

AshharAhmadKhan: Done, thanks

Aman-Mittal (approved):

Issue #301, opened 2026-08-08
Reporting: cascading parameters — loan officer depends on office, product depends on currency


Aman-Mittal:
Part of #299. Depends on #300 — parameter discovery must land first.

## Business value

Several report parameters are only meaningful relative to another. Loan officer depends on the selected office. Product often depends on currency. Centre depends on branch.

The current screen sends `rLoanOfficerId` as `undefined` explicitly (`run-report.component.ts:303`), so this has never worked. Once #300 renders the parameter, it will render as a flat list of every loan officer in the institution — which for a network of thirty branches is an unusable dropdown, and worse, lets a user pick an officer who does not work in the office they selected. The report then returns nothing, and the user has no way to know why.

## What to build

Read each parameter's dependency from the template response — the platform declares which other parameter a parameter's lookup depends on — and:

- disable a dependent control until its parent has a value
- refetch its options whenever the parent changes
- clear its value when the parent changes, so a stale selection cannot be submitted

## Two things to get right

1. **Clear the child on parent change, do not merely refetch.** Leaving the previous officer selected after switching office is the failure this issue exists to prevent, and it produces an empty report with no explanation.
2. **Handle a slow or failed lookup.** A dependent dropdown that silently stays empty looks identical to one with no options. `app-data-table` had exactly this problem and it took issue #223 to surface it — do not reintroduce the pattern here.

## Testing

- **Unit spec:** a dependent control is disabled until its parent has a value.
- **Unit spec:** changing the parent refetches options and clears the child's value.
- **Unit spec:** a failed lookup shows an error state, not an empty dropdown.
- **Mocked e2e:** select an office, assert the loan officer list is scoped; change office, assert the previous selection is gone.

## Scope

In scope: parameter dependency resolution, refetch and clearing.

Out of scope: parameter discovery itself (#300); output formats; report definition CRUD.

## Getting started

- `src/app/features/reporting/run-report.component.ts`
- Confirm how dependencies are expressed in the template response against a live instance — `npm run e2e:stack`.
- `npm test`, `npm run lint:prune`, `npm run i18n:check`, `npm run build`

PR #308, opened 2026-08-09
fix(reporting): discover report parameters dynamically

### LINKED ISSUE #300: Reporting: discover report parameters from the template endpoint instead of hardcoding three
Part of #299. Do this one first — it is the correctness fix.

## Business value

`run-report.component.ts` offers exactly three parameters — Office, From Date, To Date — regardless of what the report being run actually declares. The parameter-template endpoint that says which parameters a report needs is never called.

The platform does not reject a report run with parameters missing. **It runs it, and returns rows for the wrong scope.** So a branch manager filtering portfolio-at-risk by loan officer gets a figure covering every officer, and nothing on screen says so. A number that is quietly wrong is worse than a screen that fails, because it gets acted on.

This is a correctness bug that happens to look like a missing feature.

## What to build

On selecting a report, call the report's parameter template and render a control per declared parameter, driven by its `displayType`:

| displayType | Control |
|---|---|
| `select` | dropdown, options from the parameter's own lookup |
| `date` | date picker |
| `text` | text input |
| `none` / hidden | not rendered, value passed through |

Then pass every collected value on the run call.

**Read the run-reports signature very carefully.** `src/app/api/api/runReports.service.ts:137` is:

```ts
getRunreportsReportName(reportName, exportCSV?, parameterType?, outputType?,
                        rOfficeId?, rLoanOfficerId?, rFromDate?, rToDate?,
                        rCurrencyId?, rAccountNo?, ...)
```

Seven consecutive optional strings. The current call site (`:303`) passes them positionally, so **an argument in the wrong slot type-checks cleanly and sends the wrong value** — `officeId` silently becomes `outputType`, `fromDate` becomes `rLoanOfficerId`. Count the positions, and consider naming each argument in a comment as `journal-entries-list.component.ts:136-151` does.

## Two things to get right

1. **A parameter the report declares but the UI cannot render must not be silently dropped.** Either render something for it or tell the user the report cannot be run here. Silently omitting it reproduces the exact bug this issue exists to fix.
2. **Keep the three current parameters working.** Whatever reports people run today must keep returning the same rows.

## Testing

- **Unit spec:** given a template declaring five parameters, five controls render, one per `displayType`.
- **Unit spec:** collected values reach the correct positional argument. Assert the full argument list, not just that the call happened — position is the whole risk here.
- **Unit spec:** a report declaring only Office renders only Office.
- **Mocked e2e:** intercept the run call and assert the query string carries the selected loan officer.
- **Backend e2e** (add to `BACKEND_SPECS` in `playwright.config.ts`): run a report with two different values for a parameter and assert the row sets differ. **This is the test that proves the bug is fixed** — one that only checks the table renders would have passed all along.

## Scope

In scope: parameter discovery, rendering by `displayType`, and passing values correctly.

Out of scope: cascading parameters (#301); output formats and renderers; report definition CRUD.

## Getting started

- `src/app/features/reporting/run-report.component.ts`
- `src/app/api/api/runReports.service.ts`
- `npm run e2e:stack` for a live instance to inspect real template responses.
- `npm test`, `npm run lint:prune`, `npm run i18n:check`, `npm run build`

iamhaseebn:
## What and why

Report execution hardcoded Office and two dates even though Fineract defines each report parameter dynamically. This change discovers the selected report template, renders every supported control type, fails closed for unknown types, and sends the complete named R_* parameter map for both HTML and CSV execution.

Select lookups are isolated per field: parent-dependent lookups wait for cascading support in #301, one failed lookup does not cancel its siblings, and malformed template rows do not discard valid parameters.

Closes #300

## Verification

- `npm run lint`
- `npm run format:check`
- `npm run i18n:check`
- `npm run check:icons`
- `npm run build`
- `TZ=UTC npm test -- --watch=false --browsers=ChromeHeadless` - 853 tests passed
- `npx playwright test e2e/report-enhancements.spec.ts e2e/reporting.spec.ts e2e/functional-coverage.spec.ts --project=mocked --workers=2` - 45 tests passed
- `npx playwright test e2e/report-parameter-backend.spec.ts --project=backend --workers=1` - backend setup and regression tests passed against a real local Fineract stack
- The mocked flow verifies both `R_officeId` and `R_loanOfficerId` in the outgoing query
- The Client Listing backend flow proves that changing Office changes the returned row set
- The `Active Loans - Summary` backend flow proves cascading lookups no longer take down the parameter form

## Screenshots

A screenshot and short recording of the dynamic report parameters are attached in the evidence comment below.

## Checklist

- [x] New component or service code avoids direct browser globals and imperative third-party APIs.
- [x] I added unit, mocked e2e, and real-backend regression coverage.
- [x] UI workflow changes include suitable e2e coverage with both mocks and a real Fineract backend.
- [x] The commits are signed and include the Signed-off-by trailer.

### DIFF
--- a/src/app/features/reporting/report-execution.service.ts
+++ b/src/app/features/reporting/report-execution.service.ts
@@ -1,0 +1,182 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { HttpClient, HttpParams } from '@angular/common/http';
+import { Injectable, inject } from '@angular/core';
+import { Observable, forkJoin, map, of, switchMap } from 'rxjs';
+
+import { ConfigService } from '../../core/services/config.service';
+
+export interface ReportSelectOption {
+  readonly id: string | number;
+  readonly name: string;
+  readonly isAll?: boolean;
+}
+
+export interface ReportParameter {
+  readonly name: string;
+  readonly variable: string;
+  readonly label: string;
+  readonly displayType: string;
+  readonly formatType: string;
+  readonly defaultValue: unknown;
+  readonly selectOne: unknown;
+  readonly selectAll: unknown;
+  readonly parentParameterName: string | null;
+  readonly queryParameter: string;
+  readonly options: readonly ReportSelectOption[];
+}
+
+export interface ReportResult {
+  readonly columnHeaders?: readonly Record<string, unknown>[];
+  readonly data?: readonly Record<string, unknown>[];
+  readonly [key: string]: unknown;
+}
+
+interface GenericResultset {
+  readonly data?: readonly { readonly row?: unknown }[];
+}
+
+export type ReportParameterValues = Readonly<Record<string, string | number>>;
+
+/**
+ * The generated run-reports client exposes only a fixed positional subset of report parameters.
+ * Report definitions are dynamic, so this service deliberately uses a named query map instead.
+ */
+@Injectable({ providedIn: 'root' })
+export class ReportExecutionService {
+  private readonly http = inject(HttpClient);
+  private readonly config = inject(ConfigService);
+
+  getReportParameters(reportName: string): Observable<ReportParameter[]> {
+    const params = new HttpParams().set('R_reportListing', reportName).set('parameterType', 'true');
+
+    return this.http.get<GenericResultset>(this.reportUrl('FullParameterList'), { params }).pipe(
+      map((response) => (response.data ?? []).map((entry) => this.toReportParameter(entry.row))),
+      switchMap((parameters) => this.loadSelectOptions(parameters)),
+    );
+  }
+
+  runReport(reportName: string, values: ReportParameterValues): Observable<ReportResult> {
+    const params = this.buildRunParams(values, false, 'HTML');
+    return this.http.get<ReportResult>(this.reportUrl(reportName), { params });
+  }
+
+  downloadCsv(reportName: string, values: ReportParameterValues): Observable<string> {
+    const params = this.buildRunParams(values, true, 'CSV');
+    return this.http.get(this.reportUrl(reportName), { params, responseType: 'text' });
+  }
+
+  private loadSelectOptions(parameters: ReportParameter[]): Observable<ReportParameter[]> {
+    const requests = parameters.map((parameter) => {
+      if (parameter.displayType !== 'select') {
+        return of(parameter);
+      }
+
+      const params = new HttpParams().set('parameterType', 'true');
+      return this.http.get<GenericResultset>(this.reportUrl(parameter.name), { params }).pipe(
+        map((response) => {
+          const options = (response.data ?? []).map((entry) => this.toSelectOption(entry.row));
+          const offersAll = String(parameter.selectAll).toUpperCase() === 'Y';
+          const alreadyOffersAll = options.some((option) => String(option.id) === '-1');
+
+          return {
+            ...parameter,
+            options:
+              offersAll && !alreadyOffersAll
+                ? [...options, { id: '-1', name: '', isAll: true }]
+                : options,
+          };
+        }),
+      );
+    });
+
+    return requests.length > 0 ? forkJoin(requests) : of([]);
+  }
+
+  private toReportParameter(row: unknown): ReportParameter {
+    if (!Array.isArray(row) || row.length < 9) {
+      throw new Error('The report parameter template returned an invalid row.');
+    }
+
+    const [
+      name,
+      variable,
+      label,
+      displayType,
+      formatType,
+      defaultValue,
+      selectOne,
+      selectAll,
+      parent,
+    ] = row;
+    if (
+      typeof name !== 'string' ||
+      typeof variable !== 'string' ||
+      typeof displayType !== 'string'
+    ) {
+      throw new Error('The report parameter template omitted a required field.');
+    }
+
+    return {
+      name,
+      variable,
+      label: typeof label === 'string' && label ? label : variable,
+      displayType: displayType.toLowerCase(),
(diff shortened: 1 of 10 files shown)


iamhaseebn:
## Visual evidence

The screenshot shows the template-discovered Office and Loan Officer selectors populated from their report lookups. The short recording covers selecting both values, sending the complete named parameter query, rendering the report, pagination, and CSV download.

dynamic-report-parameters

[View the short recording](https://github.com/user-attachments/assets/004d9a80-ab79-4810-bd0c-d69a3244d57b)

src/app/features/reporting/report-execution.service.ts
```diff
@@ -0,0 +1,182 @@
...
+
+          return {
+            ...parameter,
+            options:
+              offersAll && !alreadyOffersAll
+                ? [...options, { id: '-1', name: '', isAll: true }]
+                : options,
+          };
+        }),
+      );
```
Aman-Mittal: This fetch is eager and the `forkJoin` below is all-or-nothing, so one lookup that cannot succeed takes the whole parameter form down with it.

Two of the platform's stock lookups are **cascading** — they declare a parent and their SQL cannot run until the parent has a value:

```
GET /runreports/loanOfficerIdSelectAll?parameterType=true   → 403 (data integrity / SQL error)
GET /runreports/loanProductIdSelectAll?parameterType=true   → 403
```

`loanOfficerIdSelectAll` declares parent `OfficeIdSelectOne`, `loanProductIdSelectAll` declares parent `currencyIdSelectAll`. Because `forkJoin` errors as soon as any inner observable errors, `getReportParameters` fails and the screen renders `report-parameters-error` with no fields at all.

I enumerated every Table report against a running Fineract: **20 of 59 stop rendering a parameter form entirely**, among them Active Loans – Summary, Active Loans – Details, Portfolio at Risk, Portfolio at Risk by Branch, Loans Pending Approval, Loans Awaiting Disbursal, Written-Off Loans, Rescheduled Loans, Client Loans Listing and Expected Payments By Date.

The suite stays green because `report-parameter-backend.spec.ts` exercises `Client Listing`, whose only parameter is `OfficeIdSelectOne` — the case that works.

This does not need cascading implemented (that is #301). It needs the fetch bounded: skip a parameter that declares a parent, and let a failed lookup degrade to an empty option list rather than failing its siblings.

```suggestion
      // A parameter that declares a parent cannot be fetched yet: its query filters on the
      // parent's value, and without one the platform answers 403 with a SQL error. Populating
      // these as the parent is chosen is #301; until then they render empty rather than taking
      // the rest of the form down.
      if (parameter.displayType !== 'select' || parameter.parentParameterName) {
        return of(parameter);
      }

      const params = new HttpParams().set('parameterType', 'true');
      return this.http.get<GenericResultset>(this.reportUrl(parameter.name), { params }).pipe(
        map((response) => {
          const options = (response.data ?? []).map((entry) => this.toSelectOption(entry.row));
          const offersAll = String(parameter.selectAll).toUpperCase() === 'Y';
          const alreadyOffersAll = options.some((option) => String(option.id) === '-1');

          return {
            ...parameter,
            options:
              offersAll && !alreadyOffersAll
                ? [...options, { id: '-1', name: '', isAll: true }]
                : options,
          };
        }),
        // One lookup the tenant's data cannot satisfy should cost that one field, not the report.
        catchError(() => of(parameter)),
      );
```

`catchError` needs adding to the `rxjs` import on line 21.

Worth adding a backend test over a report with a cascading parameter — `Active Loans - Summary` is the obvious one — so this failure mode is covered rather than sitting just outside the suite.

src/app/features/reporting/report-execution.service.ts
```diff
@@ -0,0 +1,182 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { HttpClient, HttpParams } from '@angular/common/http';
+import { Injectable, inject } from '@angular/core';
```
Aman-Mittal, line 21:
```suggestion
import { Observable, catchError, forkJoin, map, of, switchMap } from 'rxjs';
```
Goes with the suggestion on `loadSelectOptions` below.

src/app/features/reporting/report-execution.service.ts
```diff
@@ -0,0 +1,182 @@
...
+      );
+    });
+
+    return requests.length > 0 ? forkJoin(requests) : of([]);
+  }
+
+  private toReportParameter(row: unknown): ReportParameter {
+    if (!Array.isArray(row) || row.length < 9) {
+      throw new Error('The report parameter template returned an invalid row.');
+    }
```
Aman-Mittal: Minor, and only a hazard if a tenant ever returns a shorter row: this `throw` is inside a `map`, so a single malformed row fails the entire parameter list rather than the one field it describes.

Every row on a stock instance is 9 columns, so nothing is broken today — but skipping the row would keep the rest of the report usable if that ever stops being true.

Aman-Mittal (commented):
Reading parameters off `FullParameterList` instead of hardcoding three is the right fix, and deliberately using a named query map rather than the generated client's fixed positional parameters is exactly right for something whose shape is defined by data. The `select`/`date`/`text`/`none` switch with an explicit note for types it does not handle is a nice touch, and the added `data-testid`s make the screen testable for the first time.

I ran this against a live Fineract before commenting. The run URL shape is correct — `exportCSV=false&output-type=HTML&R_officeId=1` answers 200 with `columnHeaders` and `data` — so execution is sound.

The one thing I would not merge as it stands is the eager option fetch. Because every `select` parameter is fetched up front through a single `forkJoin`, and two of the platform's stock lookups cannot answer without their parent's value, **20 of the 59 Table reports stop rendering a parameter form altogether** — including Active Loans – Summary, Portfolio at Risk and Loans Pending Approval. Details and a suggested fix are inline; it is a couple of lines and does not require cascading to be implemented here.

Worth noting why CI does not catch it: the new backend spec covers `Client Listing`, whose only parameter is the one that works. A test over a report with a cascading parameter would pin this down.

Two smaller notes inline as well. Nothing else structural — with the option fetch bounded this is a clear improvement on what is there now.

src/app/features/reporting/report-execution.service.ts
iamhaseebn: Thanks for validating this against the stock reports. Fixed in ffa1267: parent-dependent select lookups are no longer fetched eagerly, and an independent lookup failure now degrades only that field to an empty option list. I added unit coverage plus a real-backend `Active Loans - Summary` regression against a fresh Fineract stack.

src/app/features/reporting/report-execution.service.ts
iamhaseebn: Added `catchError` in ffa1267 as part of the bounded per-field lookup handling.

src/app/features/reporting/report-execution.service.ts
iamhaseebn: Fixed in ffa1267. Invalid `FullParameterList` rows are now skipped while valid rows continue loading, with regression coverage.

Aman-Mittal (approved):
LGTM and thank you for contribution

Issue #251, opened 2026-08-07
31 hardcoded English strings on list screens ignore the language setting


Aman-Mittal:
## Business value

The application ships three locales and offers a language switcher, so a user who selects Hindi expects the interface to be in Hindi. On seventeen list screens it partly is not: the page heading, the create button, or an action tooltip stays in English regardless of the language chosen.

The effect is worse than an untranslated string usually is, because of *which* strings these are. They are the page heading and the primary action — the two pieces of text a user reads to work out where they are and what they can do. A screen whose heading is `Chart of Accounts` and whose button is `Add Ledger Account`, sitting inside an otherwise translated shell, reads as half-broken.

These also escape the normal safety net. Everything else goes through a translation key, so a missing translation shows up as a visible key like `GROUPS.CREATE_GROUP` and diagnoses itself. A hardcoded English string looks correct in every locale and is only findable by grep.

## Reproducing it

```
grep -rn 'title="[A-Z][a-z][^"]*"\|createButtonLabel="[A-Z][a-z][^"]*"\|searchPlaceholder="[A-Z][a-z][^"]*"' \
  src/app --include=*.ts | grep -v spec
```

31 sites across 17 files. By feature: accounting 11, products 5, loans 4, tasks 3, fintech 3, centers 2, and one each in security, clients and campaigns.

A sample:

```ts
// src/app/features/accounting/chart-of-accounts.component.ts:45
title="Chart of Accounts"
createButtonLabel="Add Ledger Account"

// src/app/features/loans/collateral/collateral-list.component.ts:64
title="Edit Collateral"
```

Switch the language to Hindi or Korean and those three strings stay in English.

## Describing the change

Replace each literal with a translation key, and add the key to `src/assets/i18n/en.json` if it is not already there.

`app-data-table` translates `title`, `createButtonLabel` and `searchPlaceholder` for you, so those are a straight substitution:

```ts
// before
title="Chart of Accounts"
createButtonLabel="Add Ledger Account"

// after
title="ACCOUNTING.CHART_OF_ACCOUNTS"
createButtonLabel="ACCOUNTING.ADD_LEDGER_ACCOUNT"
```

The `title=` attributes on `ion-button` are native tooltips rather than a data-table input, so they need the pipe and a binding:

```ts
// before
<ion-button title="Edit Collateral" (click)="onEdit(row)">

// after
[...]

PR #321, opened 2026-08-09
fix(accounting): use translation keys for list screen titles and tooltips

yasumorishima:
## What and why

The five accounting list screens passed English literals to `app-data-table`'s `title` / `createButtonLabel` inputs and to `ion-button` `title` tooltips, so those strings stayed in English whichever language was selected. As #251 notes, this is worse than an ordinary missing translation: a missing key renders as a visible key and diagnoses itself, whereas a hardcoded string looks correct in every locale.

This replaces all **11 sites in `src/app/features/accounting`** with translation keys. #251 says one PR per feature directory; accounting is the largest at 11 sites.

Part of #251.

## The 11 replacements

| File | Was | Now | Key |
| --- | --- | --- | --- |
| `accounting-closures-list` | `title="Accounting Closures"` | `nav.accountingClosures` | existing |
| `accounting-closures-list` | `createButtonLabel="Close Period"` | `ACCOUNTING_CLOSURES.CREATE` | existing |
| `accounting-closures-list` | `title="Re-open Period"` | `ACCOUNTING_CLOSURES.REOPEN` | **new** |
| `chart-of-accounts` | `title="Chart of Accounts"` | `nav.chartOfAccounts` | existing |
| `chart-of-accounts` | `createButtonLabel="Add Ledger Account"` | `ACCOUNTING.ADD_LEDGER_ACCOUNT` | **new** |
| `chart-of-accounts` | `title="Edit Account"` | `ACCOUNTING.EDIT_ACCOUNT` | **new** |
| `journal-entries-list` | `title="Journal Entries"` | `nav.journalEntries` | existing |
| `journal-entries-list` | `createButtonLabel="Add Journal Entry"` | `JOURNAL_ENTRIES.CREATE` | existing |
| `financial-activity-mappings-list` | `title="Financial Activity Mappings"` | `nav.financialActivityMappings` | existing |
| `financial-activity-mappings-list` | `createButtonLabel="Define Mapping"` | `ACCOUNTING.DEFINE_MAPPING` | **new** |
| `charges-list` | `title="Edit Charge"` | `CHARGES.EDIT` | existing |

**No visible English text changes.** Every key resolves to the exact string it replaced, so in English this is a no-op and only the other locales change.

## Two decisions worth a reviewer's eye

**1. List screen titles reuse `nav.*` rather than new `ACCOUNTING.*` keys.** #251 sketched `title="ACCOUNTING.CHART_OF_ACCOUNTS"`, but 45 existing list screens already pass their `nav.*` entry as the data-table title (`nav.groups`, `nav.users`, `nav.loanProducts`, …), and `nav.chartOfAccounts` / `nav.journalEntries` / `nav.accountingClosures` / `nav.financialActivityMappings` already exist with byte-identical English. Reusing them follows both the issue's "reuse beats a near-duplicate" guidance and the dominant convention. Happy to switch to a new namespace if you prefer.

**2. `ADD_LEDGER_ACCOUNT` / `EDIT_ACCOUNT` are new even though `CREATE_GL_ACCOUNT` / `EDIT_GL_ACCOUNT` exist.** Those read "Create GL Account" / "Edit GL Account" and title the `gl-account-form` dialog these buttons open, so collapsing into them looked like the tighter catalogue. It would have broken a test: `e2e/functional-coverage.spec.ts:839` drives this screen with `getByRole('button', { name: /Add Ledger Account/i })`. Keeping the English byte-identical is what makes this change safe. If you do want the wording unified, the e2e selector has to change in the same PR.

## Verification

Run on a fork of this repository at the same commit — **all 18 jobs green**:

- **CI (12/12)**: Dependency Integrity, TS/ESLint, HTML Lint, Format Check, Unit Tests, Production Build, Translation Check, License Compliance, Apache RAT Scan, Security Audit, API Client Drift Check, Build Summary.
- **E2E (6/6)**: all four mocked shards, the mocked report, and **E2E (real Fineract)** against the docker-compose backend.

Additional checks made while preparing this:

- `node scripts/check-translations.mjs` → `Checked 1485 referenced keys against 2014 keys in en.json. ✓ No missing keys.` This is a real gate here, not a formality: the checker's patterns match the `title="KEY"` / `createButtonLabel="KEY"` attribute forms, so every key introduced above is statically validated.
- The grep in #251 drops from 31 sites to 20 repo-wide, and to **0** under `src/app/features/accounting`.
- Every existing assertion that touches these strings was checked before changing them. The e2e suite drives all five screens by their English text (`feature-coverage.spec.ts:414` asserts the card title; `functional-coverage.spec.ts:785/816/839` click the create buttons by name) and each still matches because the keys resolve to the identical string. No unit spec under `src/app` asserts these literals.
- The three converted `title=` → `[title]=` bindings use `| appTranslate` with `TranslatePipe` from `src/app/core/adapters`, per `DOCS/ADAPTERS.md`, and their attribute order matches existing prettier-stable elements of the same shape (`groups-list` L85-90, `email-campaigns-list` L115-119). `Format Check` passing confirms `prettier` is a no-op on this diff.

## Screenshots

Not applicable — no visual change in English. The change is only observable by switching the language, where these strings now translate instead of staying English.

## Checklist

- [x] I added or updated tests appropriate to this change, or explained why tests were not needed — there is no new behaviour and no visible string change, so the existing e2e specs that drive these screens by their English text are the regression test, and `check-translations` covers the new keys.
- [x] UI workflow changes include suitable e2e coverage — not applicable, no workflow change.
- [ ] **Commits are signed.** The commit *is* SSH-signed, but the public key is not yet registered as a signing key on my GitHub account, so it currently displays as `Unverified` (`GET /commits/:sha` reports `reason: "unknown_key"`, i.e. GitHub sees a valid signature it cannot attribute). Registering the key makes it verify retroactively with no rewrite of this branch, and I will do that shortly. Flagging it rather than ticking the box.

### DIFF
--- a/src/app/features/accounting/accounting-closures-list.component.ts
+++ b/src/app/features/accounting/accounting-closures-list.component.ts
@@ -23,6 +23,7 @@ import { Router } from '@angular/router';
 import { DatePipe, NgClass } from '@angular/common';
 import { DataTableComponent, ColumnDef, CellTemplateDirective } from '../../shared';
 import { AccountingClosureService, GetGlClosureResponse } from '../../api';
+import { TranslatePipe } from '../../core/adapters';
 import { IonButton, IonIcon } from '@ionic/angular/standalone';
 
 /**
@@ -37,6 +38,7 @@ import { IonButton, IonIcon } from '@ionic/angular/standalone';
     TranslateModule,
     DataTableComponent,
     CellTemplateDirective,
+    TranslatePipe,
     DatePipe,
     NgClass,
     IonIcon,
@@ -44,9 +46,9 @@ import { IonButton, IonIcon } from '@ionic/angular/standalone';
   ],
   template: `
     <app-data-table
-      title="Accounting Closures"
+      title="nav.accountingClosures"
       helpTextKey="HELP.ACCOUNTING_CLOSURES_DESC"
-      createButtonLabel="Close Period"
+      createButtonLabel="ACCOUNTING_CLOSURES.CREATE"
       [columns]="columns"
       [data]="closures()"
       [localLogic]="true"
@@ -67,7 +69,7 @@ import { IonButton, IonIcon } from '@ionic/angular/standalone';
         <ion-button
           fill="clear"
           color="danger"
-          title="Re-open Period"
+          [title]="'ACCOUNTING_CLOSURES.REOPEN' | appTranslate"
           (click)="onDeleteClosure(closure)"
         >
           <ion-icon name="lock-open-outline"></ion-icon>
--- a/src/app/features/accounting/charges/charges-list.component.ts
+++ b/src/app/features/accounting/charges/charges-list.component.ts
@@ -24,6 +24,7 @@ import { TranslateModule } from '@ngx-translate/core';
 import { CurrencyPipe, DecimalPipe } from '@angular/common';
 import { DataTableComponent, ColumnDef, CellTemplateDirective } from '../../../shared';
 import { ChargesService, ChargeData } from '../../../api';
+import { TranslatePipe } from '../../../core/adapters';
 import { IonButton, IonIcon } from '@ionic/angular/standalone';
 
 /**
@@ -38,6 +39,7 @@ import { IonButton, IonIcon } from '@ionic/angular/standalone';
     TranslateModule,
     DataTableComponent,
     CellTemplateDirective,
+    TranslatePipe,
     DecimalPipe,
     CurrencyPipe,
     IonIcon,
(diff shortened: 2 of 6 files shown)


Aman-Mittal:
@yasumorishima please upload your signing key and remove claude co-authored in commit

This user has not yet uploaded their public signing key.
[Learn about vigilant mode](https://docs.github.com/github/authenticating-to-github/displaying-verification-statuses-for-all-of-your-commits)

yasumorishima:
Thanks for the review — both are done.

- Uploaded my SSH signing key, so `91d2f84` now shows as Verified.
- Amended the commit message to drop the co-author trailer.

The amend touched the message only: the tree is unchanged at `a29f6697`, so there is no code difference from what you looked at. The workflows went back to `action_required` after the force-push and need approval to run again.

PR #323, opened 2026-08-10
fix: reject maker-checker tasks without deleting audit entries

### LINKED ISSUE #280: Rejecting a maker-checker task deletes the record instead of recording the rejection
## Business value

The checker inbox approves and rejects through two different verbs, and only one of them is right.

`src/app/features/tasks/checker-inbox/checker-inbox.component.ts:143` (approve) does the correct thing:

```ts
this.makerCheckerService.postMakercheckersAuditId(task['id'] as number, 'approve')
```

`:156` (reject) does not:

```ts
this.makerCheckerService.deleteMakercheckersAuditId(task['id'] as number)
```

The generated client documents the two endpoints as distinct operations:

```
post   /v1/makercheckers/{auditId}   Approve Maker Checker Entry | Reject Maker Checker Entry
delete /v1/makercheckers/{auditId}   Delete Maker Checker Entry
```

Verified against a live backend — `command=reject` is a recognised command, and the platform enumerates its accepted values when you send a bad one:

```
POST /v1/makercheckers/1?command=totallyNotACommand   400  "unsupported value of: totallyNotACommand"
POST /v1/makercheckers/1?command=reject               404  "Audit with identifier 1 is Not Awaiting Approval"
```

The second response is the proof: `reject` passed command validation and reached the audit lookup.

**Why this matters more than a wrong verb usually would.** Maker-checker exists to produce an audit trail: who proposed a change, who reviewed it, and what they decided. Approving records an approval. Rejecting is supposed to record a rejection against the same entry. Deleting removes the entry, so the record of the proposal *and* the decision disappears together. An auditor asking "who rejected this loan write-off, and when?" has nothing to look at, and the UI reports "Task rejected successfully" either way — so nobody discovers the loss until it is needed.

For an institution using four-eyes approval to satisfy a regulator, a rejection path that destroys evidence is a control failure, not a cosmetic bug.

## Reproducing it

```
grep -n "postMakercheckersAuditId\|deleteMakercheckersAuditId" \
  src/app/features/tasks/checker-inbox/checker-inbox.component.ts
```

Two call sites, two different endpoints, one operation.

## Describing the change

```ts
// before
this.makerCheckerService.deleteMakercheckersAuditId(task['id'] as number).subscribe({ ... });

// after
this.makerCheckerService.postMakercheckersAuditId(task['id'] as number, 'reject').subscribe({ ... });
```

That is the fix. Two further things in the same handler are worth doing while you are there, and both are small:

1. **`confirm()`.** `:155` uses the browser's native confirm box. It cannot be translated, cannot be styled, and cannot be reached from the e2e suite. Replace it with `DialogService.confirm({ title, message, destructive: true })` — the pattern is in `src/app/features/groups/tabs/group-notes-list.component.ts:166`. The same substitution is tracked elsewhere for other screens (#229–#232); this call site is not in any of them.
2. **Hardcoded English.** `'Task rejected successfully'` and `'Failed to reject task'` bypass translation. Add keys alongside the existing ones.

Do all three in one PR; they are the same twelve lines.

## Testing

Unlike most issues here there is no cucumber scenario to lift — the platform's acceptance suite exercises maker-checker through the API rather than as a user-facing flow. So the test has to assert the contract directly:

**A unit spec** in a new `checker-inbox.component.spec.ts`. There is currently none, and the assertion is exactly the thing that was wrong:

```ts
it('rejects through the reject command, not by deleting the entry', async () => {
  adapters.overlay.nextModalResult = true;          // confirm
  await component.onReject({ id: 7 });

  expect(serviceSpy.postMakercheckersAuditId).toHaveBeenCalledWith(7, 'reject');
  expect(serviceSpy.deleteMakercheckersAuditId).not.toHaveBeenCalled();
});
```

The negative assertion matters. A spec that only checks the POST would still pass if both calls fired.

Also cover: cancelling the confirmation issues no call at all.

Use `provideFakeAdapters()` from `src/app/testing/adapters.ts` and `provideTranslateTesting()` from `src/app/testing/i18n-testing.ts` — importing `@ngx-translate/core` directly in a spec is banned by lint (ADR-0003) and CI will fail on it.

**A mocked e2e** is welcome: intercept `POST /makercheckers/*`, assert the `command` search parameter is `reject`, and fail the test if any `DELETE` reaches that path. See the `Probe` pattern in `e2e/client-servicing-gaps.spec.ts`.

## Scope

In scope: the reject verb, the confirm dialog, and the two hardcoded strings in this handler.

Out of scope: everything else about the checker inbox — it also has no search, no filtering and no bulk approve/reject, which is a much larger piece of work and deserves its own issue.

## Getting started

- `src/app/features/tasks/checker-inbox/checker-inbox.component.ts`
- Confirmation pattern: `src/app/features/groups/tabs/group-notes-list.component.ts:165-183`
- Spec harness: `src/app/testing/adapters.ts`, `src/app/testing/i18n-testing.ts`
- `npm test` and `npm run lint:prune` must pass. Run `lint:prune` *after* writing the spec, not before — it prunes suppressions and will fail on a spec that imports the translate module directly.

chengzhipengczp:
## What and why

Use the maker-checker reject command endpoint instead of deleting audit entries.

The rejection handler previously called the DELETE endpoint, which removed the
maker-checker audit entry instead of recording the rejection.

Closes #280

## Verification

Added a regression test covering the rejection behavior.

Verified that:
- reject uses `postMakercheckersAuditId(id, 'reject')`
- delete endpoint is not called
- success notification is shown
- inbox refreshes after rejection

Tests passed locally:
- npm test
- npm run lint

### DIFF
--- a/src/app/features/tasks/checker-inbox/checker-inbox.component.spec.ts
+++ b/src/app/features/tasks/checker-inbox/checker-inbox.component.spec.ts
@@ -2,7 +2,8 @@
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
+ * regarding copyright ownership.
+ * The ASF licenses this file
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


chengzhipengczp:
Thanks for reviewing.

This PR fixes the rejection flow by using the maker-checker reject endpoint instead of deleting the audit entry.

Local verification:
- npm test
- npm run lint

Aman-Mittal:
@chengzhipengczp plz sign your commits with a gpg key and upload it in your settings.

chengzhipengczp:
Hi maintainers,

Thanks for reviewing this PR.

I investigated the maker-checker rejection flow and found that the UI was calling the DELETE endpoint, which removed the audit entry instead of recording a rejection.

This PR updates the flow to use the reject command endpoint and adds a regression test to prevent the issue from coming back.

The commit is GPG signed as requested, and local verification passed with:
- npm test
- npm run lint

Please let me know if any changes are needed.

Thanks!

Aman-Mittal:
@chengzhipengczp CI checks are failing

chengzhipengczp:
Security Audit is failing due to existing dependency vulnerabilities
(tar, vite, ws, socket.io-adapter, etc.).

These vulnerabilities are from the existing dependency tree and are unrelated to this PR.

The maker-checker rejection change has been verified locally:
- Unit test: passed
- Production build: passed
- ESLint: passed
- Format check: passed

Aman-Mittal:
@chengzhipengczp Can you also run this too?

fix available via `npm audit fix`
node_modules/ws
  socket.io-adapter  2.5.2 - 2.5.6
  Depends on vulnerable versions of ws
  node_modules/socket.io-adapter

chengzhipengczp:
Hi Aman, I ran `npm audit fix` as suggested.

It reduced the reported vulnerabilities, but it updated a large number of dependencies (including Angular devkit packages) and changed package-lock.json significantly.

Since this PR is only for the maker-checker rejection fix, I prefer to avoid mixing dependency upgrades with the functional change.

I can prepare a separate dependency update PR if needed.

Aman-Mittal:
> Hi Aman, I ran `npm audit fix` as suggested.
> 
> It reduced the reported vulnerabilities, but it updated a large number of dependencies (including Angular devkit packages) and changed package-lock.json significantly.
> 
> Since this PR is only for the maker-checker rejection fix, I prefer to avoid mixing dependency upgrades with the functional change.
> 
> I can prepare a separate dependency update PR if needed.

Then you can generate a separate PR for that first i will merge this after that.

chengzhipengczp:
Thanks, I will keep this PR focused on the maker-checker rejection fix.

I will create a separate PR for the npm audit dependency updates.

Aman-Mittal (approved):

Aman-Mittal:
@chengzhipengczp Can you recreate PR against Main? it was at wrong branch

Issue #256, opened 2026-08-07
The admin, campaigns and interop features have 16 components and not one unit test


Aman-Mittal:
## Business value

The project runs 143 component specs against 252 components, and CI runs them on every pull request — so unit tests here are load-bearing, not decoration. Three feature directories are not covered at all:

| Feature | Components | Specs |
| --- | --- | --- |
| `admin` | 6 | 0 |
| `campaigns` | 5 | 0 |
| `interop` | 5 | 0 |

What makes this worth fixing rather than just a number: these are the features least likely to be caught by anything else. The e2e suite reaches roughly 17% of routes and does not go near them, and they are not on a path anyone walks daily, so a regression here surfaces when someone tries to run a batch job or send a campaign — which is exactly the moment they need it to work.

There is also a specific class of bug that only a test catches here. Angular 22 makes `OnPush` the default, so a plain field assigned from an HTTP callback holds the right value while the view keeps showing the old one. `scripts/audit-async-state.mjs` counts those. A component with no spec has nothing standing between it and that failure, and the symptom — a dropdown that renders empty — looks like a backend problem.

## The components

```
admin/batch-operations/batch-operations.component.ts
admin/cob-tools/cob-tools.component.ts
admin/external-events/external-events.component.ts
admin/inline-job/inline-job.component.ts
admin/progressive-loan/progressive-loan-model.component.ts
admin/wc-cob-tools/wc-cob-tools.component.ts

campaigns/email-campaigns/email-campaign-form.component.ts
campaigns/email-campaigns/email-campaigns-list.component.ts
campaigns/email-messages/email-messages.component.ts
campaigns/sms-campaigns/sms-campaign-form.component.ts
campaigns/sms-campaigns/sms-campaigns-list.component.ts

interop/interop-account-view.component.ts
interop/interop-party-lookup.component.ts
interop/interop-quotes.component.ts
interop/interop-transfers.component.ts
interop/interop-health.component.ts
```

Verify with:

```
find src/app/features/admin src/app/features/campaigns src/app/features/interop \
  -name '*.component.ts' ! -name '*.spec.ts' | wc -l   # 16
find src/app/features/admin src/app/features/campaigns src/app/features/interop \
  -name '*.component.spec.ts' | wc -l                  # 0
```

## Describing the change

Add a `*.component.spec.ts` beside each component. **One component per pull request** — a spec is only useful if someone reads it, and sixteen at once will not be read.

Worth asserting, in rough order of value:

[...]

PR #332, opened 2026-08-11
test(interop): cover health check states

YinkaMetrics:
## What and why

Add focused unit coverage for the Interop Health component's pending, success, and error states. The error case verifies that the health-check action is re-enabled after a failed request.

Part of #256.

## Verification

- 3 focused tests passed
- 899 full tests passed


Aman-Mittal:
Merged — this is on `main` as f166c094, with you as the author. It went in through the rollup in #389 rather than from this branch, because the commit here was unsigned and the repository requires signatures; cherry-picking and re-signing keeps the authorship and only changes the committer. Nothing about the change itself was altered.

Closing this as merged rather than rejected. Thank you — #256 asks for one component per PR and this is exactly that shape.

Issue #257, opened 2026-08-07
18 sites rebuild a Date just to reformat it, when formatArrayDate already does exactly that


Aman-Mittal:
## Business value

Fineract returns dates as `[year, month, day]` with a 1-based month. Eighteen places convert one to a `YYYY-MM-DD` string by constructing a `Date` and formatting it back:

```ts
toIsoDate(new Date(actDateArray[0], actDateArray[1] - 1, actDateArray[2]))
```

`src/app/core/utils/date-formatter.ts` already exports a function that does this directly:

```ts
formatArrayDate(actDateArray)
```

The two produce identical output for a valid array — worth confirming, since the round trip through `Date` is not obviously lossless, and it is:

```
[2026, 1, 5]   → 2026-01-05   both
[2026, 12, 31] → 2026-12-31   both
[1999, 2, 28]  → 1999-02-28   both
```

The value in fixing it is the `- 1`. Eighteen hand-written copies of a month-offset conversion is eighteen chances to write `[1]` instead of `[1] - 1`, and an off-by-one month in a banking application is not a cosmetic bug — it moves an activation date, a disbursement date, or a closure date by a month, and the result is a plausible date rather than an obviously broken one, so nothing downstream flags it. The helper is written once and has a test.

It is also a tidier read: `formatArrayDate(dates)` says what it does; `toIsoDate(new Date(d[0], d[1] - 1, d[2]))` makes the reader verify the arithmetic.

## Reproducing it

```
grep -rn 'toIsoDate(new Date(.*\[0\]' src/app --include=*.ts | grep -v spec   # 18
grep -rl 'formatArrayDate' src/app --include=*.ts | grep -v spec | wc -l      # 15 files already use it
```

## Describing the change

Replace the round trip with the helper:

```ts
// before
const actDateArray = clientData.activationDate as unknown as number[];
if (actDateArray) {
  this.activationDate.set(toIsoDate(new Date(actDateArray[0], actDateArray[1] - 1, actDateArray[2])));
}

// after
this.activationDate.set(formatArrayDate(clientData.activationDate));
```

**Two things to be careful about — please read these before starting.**

1. **Only the `toIsoDate(new Date(...))` sites.** There are 13 further sites that do
   `new Date(d[0], d[1] - 1, d[2]).toLocaleDateString()`. Those are **not** the same thing:
   `toLocaleDateString()` formats for the browser's locale (`8/7/2026`), while `formatArrayDate`
[...]

PR #333, opened 2026-08-11
refactor(organization): use formatArrayDate helper

YinkaMetrics:
## What and why

Replace the manual array-to-`Date` round trip in the Organization office and staff forms with the existing `formatArrayDate` helper. Existing missing-date guards remain in place, and the component specs cover the formatted office opening and staff joining dates.

Part of #257.

## Verification

- 11 focused tests passed
- 897 full tests passed


Aman-Mittal:
Merged — this is on `main` as 301ba8e5, with you as the author. It went in through the rollup in #389 rather than from this branch, because the commit here was unsigned and the repository requires signatures; cherry-picking and re-signing keeps the authorship and only changes the committer. The change itself is untouched.

Closing this as merged rather than rejected. Thank you.

PR #337, opened 2026-08-12
ci: add signed commit verification check

Abhishek6122008:
Closes #336

Picking this one up — initial commit, happy to take feedback before I go further.

Ports the signed-commit check from `apache/fineract` (`verify-commits.yml` + `scripts/verify-signed-commits.sh`) so contributor expectations match between the two repos.

**What's here**
- `.github/workflows/signed-commit.yml` — runs on `pull_request` (opened/synchronize/reopened), checks out with full history, verifies the PR commit range in `--strict` mode so unsigned commits fail the job.
- `scripts/verify-signed-commits.sh` — walks `merge-base..head`, prints a per-commit signed/unsigned line, and emits `::error title=Unsigned Commit::` annotations so the offending commits are called out directly in the job log.

**Notes**
- Only `%G?` == `N` (no signature) fails. A runner has no contributor public keys, so `E`/`U` would flag every *signed* commit as broken — GitHub's own verification covers key validity. Same behaviour as the fineract script.
- Default `--base-ref` is `origin/main` (fineract's is `origin/develop`).
- Pinned `actions/checkout` to the same SHA already used by `zizmor.yml`, with `persist-credentials: false`.
- The docs half of the acceptance criteria already exists — [CONTRIBUTING.md#commit-signing](https://github.com/apache/fineract-backoffice-ui/blob/main/CONTRIBUTING.md#commit-signing) covers `user.signingkey` / `commit.gpgsign` and retroactive signing, so the script points there instead of duplicating it. Lemme know if you'd rather have a separate `docs/ci.md`.

Verified locally against a range containing a known unsigned commit — it's correctly identified and `--strict` exits 1.

cc @Aman-Mittal

PR #339, opened 2026-08-14
feat(router): set a translated document title per route

### LINKED ISSUE #250: Every browser tab reads "Fineract Backoffice UI" — none of the 322 routes sets a title
## Business value

Back-office work is multi-tab work. An officer keeps a client open in one tab, a loan account in another and the chart of accounts in a third, and switches between them all day.

Today every one of those tabs is labelled `Fineract Backoffice UI`, because that is the static `<title>` in `src/index.html:23` and nothing ever replaces it. The tab strip carries no information at all, so finding the right tab means clicking through them. Browser history is equally useless: every entry has the same name, and so does every bookmark the user makes.

It also matters for screen-reader users. The document title is announced on navigation and is the primary way a non-visual user learns that the page changed. In a single-page application that never updates it, navigation is silent.

Angular has first-class support for this and the application simply does not use it: `Route.title` is applied by the router's built-in `TitleStrategy` with no wiring at all.

## Reproducing it

```
grep -h "path: '" src/app/app.routes.ts src/app/features/*/*.routes.ts | wc -l   # 322
grep -c 'title:' src/app/app.routes.ts src/app/features/*/*.routes.ts           # 0 in every file
```

Then open the app and watch the tab while navigating from the dashboard to a client to a loan. It never changes.

## Describing the change

Add a `title` to each route. Routes are already split per feature — `src/app/features/<feature>/<feature>.routes.ts`, 23 files — so this is naturally one small pull request per feature, and **a first contribution should take a single feature file rather than all of them.**

```ts
{
  path: 'view/:id',
  title: 'GROUPS.GROUP_DETAILS',
  loadComponent: () => import('./group-view.component').then((m) => m.GroupViewComponent),
},
```

Two things to decide while doing it, and either is a reasonable first PR:

1. **Translation.** A raw English string in `title` would be the only user-visible text in the app that cannot be translated. The tidy answer is a custom `TitleStrategy` that resolves the route's `title` as a translation key through the `I18N` adapter (`src/app/core/adapters`) and appends a suffix, so tabs read `Groups · Fineract`. Whoever takes the first feature is welcome to add that strategy in the same PR — or to add plain keys first and leave the strategy to a follow-up, provided the keys are real entries in `src/assets/i18n/en.json`.
2. **Detail routes.** `/clients/view/:id` cannot name the client from a static string. Start with the entity name (`CLIENTS.CLIENT_DETAILS`); resolving the record's own name into the title is a separate, larger piece of work and is explicitly out of scope here.

## Scope

In scope: a `title` on the routes of one or more feature route files, the matching keys in `en.json`, and optionally the `TitleStrategy` that translates them.

Out of scope: putting record names (client name, loan account number) into titles, and any change to `index.html`.

## Getting started

- Routes live in `src/app/features/<feature>/<feature>.routes.ts`.
- Translation keys go in `src/assets/i18n/en.json`, under the feature's existing section.
- `DOCS/ADAPTERS.md` covers the `I18N` adapter, which is how this codebase reaches translation.
- `npm run build` and `npm run lint` must pass. If you add a `TitleStrategy`, a unit spec for it is welcome.

opensource-joe:
## What and why

Every browser tab read "Fineract Backoffice UI". Across 325 route declarations in 23 files, none set `Route.title`, and the app registered no `TitleStrategy`, so the static title in `src/index.html` was never replaced. The tab strip carried no information, and history entries and bookmarks were equally undifferentiated.

It also matters without a screen. The document title is announced on navigation and is the primary way a non-visual user learns the page changed, so in a single-page app that never updates it, navigation is silent.

`TranslatedTitleStrategy` treats `Route.title` as a translation key rather than a phrase, resolves it through the I18N adapter and appends the short app name, giving `Groups · Fineract`. Writing the phrase directly would have made the tab the only user-visible string in the app that cannot be translated.

Closes #250

## On scope, since the issue invites a choice

The issue offers two paths: plain keys first with the strategy as a follow-up, or the strategy in the same PR. This takes the second, because the strategy is the part that decides what every later PR looks like, and it is easier to argue about once it exists.

It also uses inheritance rather than titling all 325 routes. A route without a `title` inherits its nearest titled ancestor, which is Angular's own `buildTitle` behaviour. So titling the 25 sections in `app.routes.ts` gives every page beneath them a sensible tab immediately, and a feature can then refine its own routes without touching anything outside its own file. `groups.routes.ts` is fully refined here as the worked example of that second step.

That leaves the per-feature refinement of the remaining 22 route files as exactly the small, self-contained tasks the issue describes, which seemed more useful to you than one 325-line diff from me. Happy to take them on if you would rather, or to leave them as newcomer tasks.

Four sections are deliberately not titled, because no existing key names them and I did not want to invent page names: `admin`, `spm`, `fineract-mfe`, and the `**` route (which only redirects). They fall back to `Fineract`, which is still better than today.

## Keys

No new page names were invented. The 25 sections reuse the sidebar's own `nav.*` keys, which are already the canonical names for these pages, or an existing section `TITLE` where there is no `nav` entry (`MEETINGS.TITLE`, `CALENDARS.TITLE`, `COLLECTION_SHEET.TITLE`, `SEARCH.TITLE`, `FORGOT_PASSWORD.TITLE`, `PROFILE.TITLE`, `login.title`).

Two keys are added: `app.shortTitle` ("Fineract") for the suffix, since `app.title` is "Fineract Backoffice UI" and repeating that on every tab defeats the purpose; and `GROUPS.GROUP_DETAILS` for the group view page, which had no name.

## One small addition worth calling out

`scripts/check-translations.mjs` now recognises the route form, `title: 'nav.groups'`. Without it a typo in a route key is caught by nothing and reaches the tab as a raw key. The pattern requires the dotted-identifier key shape, so it does not match ordinary `title:` properties holding a phrase. It raises the referenced-key count from 1,439 to 1,514 and reports no missing keys.

## Verification

Run in a clean container (node 22.23.2), all passing: `lint:prune` (with `eslint-suppressions.json` unchanged), `format:check`, `eslint "src/**/*.html"`, `i18n:check`, `check:icons`, `check-license.sh`, `build`, and `npm test` at 900 of 900, up from 896.

The four new specs drive the real Angular router through `RouterTestingHarness`, the real `TitleStrategy` base class and the real `Title` service, with only the I18N adapter faked, so they cover key resolution, ancestor inheritance, the no-title fallback and re-application on language change.

I checked they are not vacuous. Removing the `effect` that watches the language fails exactly the language test and nothing else. Passing the key through untranslated and dropping the suffix fails the two resolution tests. Restoring the implementation returns all 900 to green.

What that does not cover, stated plainly: nothing here exercises the `app.config.ts` provider registration in a real browser. I had a headless check part-written and stopped it, so the wiring itself is verified by reading rather than by running. It is three lines next to `provideRouter`, but it is the one part a reviewer should not take on trust from the test results alone.

## Screenshots

Not applicable in the usual sense, since nothing inside the page changes. The visible difference is the browser tab, which now reads for example `Groups · Fineract` on `/groups` and `Create Group · Fineract` on `/groups/create`, instead of `Fineract Backoffice UI` everywhere.

## Checklist

- [x] New component or service code uses the adapter boundary in `src/app/core/adapters/` instead of direct browser globals or imperative third-party APIs. The strategy resolves keys through the `I18N` token and sets the title through Angular's `Title` service rather than touching `document` directly.
- [x] User-facing strings use translation keys. That is the substance of the change.
- [x] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. No workflow changes: no route path, guard, redirect or component was touched, only the `title` property and the strategy that reads it.
- [x] Commits are signed.

### DIFF
--- a/scripts/check-translations.mjs
+++ b/scripts/check-translations.mjs
@@ -83,6 +83,10 @@ const PATTERNS = [
     String.raw`\[(?:title|helpTextKey|label|placeholder|createButtonLabel)\]\s*=\s*['"]'${KEY}'['"]`,
     'g',
   ),
+  // Route definitions: `title: 'nav.groups'` in a *.routes.ts Routes array. Resolved as a
+  // key by TranslatedTitleStrategy, so a typo here is a missing key like any other. The KEY
+  // shape (dotted identifier) keeps this off ordinary `title:` properties holding a phrase.
+  new RegExp(String.raw`\btitle:\s*['"]${KEY}['"]`, 'g'),
 ];
 
 function referencedKeys(files) {
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -27,7 +27,7 @@ import {
   ApplicationConfig,
   ErrorHandler,
 } from '@angular/core';
-import { provideRouter } from '@angular/router';
+import { TitleStrategy, provideRouter } from '@angular/router';
 import { provideHttpClient, withInterceptors } from '@angular/common/http';
 import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
 import { TranslateModule } from '@ngx-translate/core';
@@ -44,6 +44,7 @@ import { ConfigService } from './core/services/config.service';
 import { provideIonicAngular } from '@ionic/angular/standalone';
 
 import { BASE_PATH } from './api/variables';
+import { TranslatedTitleStrategy } from './core/router/translated-title.strategy';
 
 /**
  * Factory function to load configuration before app bootstrap
@@ -74,6 +75,8 @@ export const appConfig: ApplicationConfig = {
     ...(isDevMode() ? [provideCheckNoChangesConfig({ interval: 500, exhaustive: true })] : []),
     { provide: ErrorHandler, useClass: GlobalErrorHandler },
     provideRouter(routes),
+    // Routes carry a translation key in `title`; this resolves it. See the strategy's own docs.
+    { provide: TitleStrategy, useClass: TranslatedTitleStrategy },
     // Order is the chain order, outermost first. `retryInterceptor` is last, and so closest
     // to the backend, deliberately: `loadingInterceptor` outside it counts one logical request
     // instead of flickering the progress bar per attempt, `errorInterceptor` toasts only once
--- a/src/app/app.routes.ts
+++ b/src/app/app.routes.ts
@@ -25,6 +25,7 @@ import { loadRemoteModule } from '@angular-architects/native-federation';
 export const routes: Routes = [
   {
     path: 'login',
+    title: 'login.title',
     loadComponent: () => import('./features/login/login.component').then((m) => m.LoginComponent),
   },
   {
@@ -39,6 +40,7 @@ export const routes: Routes = [
       },
       {
         path: 'dashboard',
+        title: 'nav.dashboard',
         loadComponent: () =>
           import('./features/dashboard/system-status.component').then(
             (m) => m.SystemStatusComponent,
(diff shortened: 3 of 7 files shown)


Aman-Mittal:
@opensource-joe You can amend your commit by removing claude co-authored and you are good to go

Aman-Mittal:
@opensource-joe Also add some video/screenshots

opensource-joe:
@Aman-Mittal Screenshots below. Every tab here is a real tab in one Chromium window, driven by this repo's own mocked-backend pattern (the `mocked` Playwright project), so none of it needs a Fineract instance to reproduce.

Tab strip before and after

Row 1 is `main`: six different sections, six identical tabs. Row 2 is this branch. Row 3 is the same tabs after switching the header language selector to Korean, with no navigation in between, which is the re-apply on language change the commit message describes.

That runtime switch, as title strings:

```
/clients:    "Clients · Fineract"    -> "고객 · Fineract"    (url unchanged: true)
/groups:     "Groups · Fineract"     -> "그룹 · Fineract"    (url unchanged: true)
/accounting: "Accounting · Fineract" -> "회계 · Fineract"    (url unchanged: true)
```

One detail worth raising rather than leaving for you to spot: `app.shortTitle` exists only in `en.json`, so the `· Fineract` suffix surviving in the Korean tabs is `setFallbackLang('en')` doing its job, not an untranslated key reaching the tab. If you would rather that suffix be localised too, say the word and I will add the key to `hi.json` and `ko.json`.

<details>
<summary>Full-window captures</summary>

Before, on `main`:

Before, on main

After, on this branch:

After, on this branch

After, language switched to Korean at runtime:

Korean titles after a runtime language switch

</details>

Working on the commit amend and the signature check next.

opensource-joe:
@Aman-Mittal Amended. The Claude co-author trailer is gone from this branch and from #338, and both commits are now GPG signed. The trees are byte-identical to before, so this is a message and signature change only.

On the check itself, one thing worth flagging because it will hit every SSH signer and not just me.

`scripts/verify-signed-commits.sh` fails only on `%G?` == `N`, and the comment inside it is explicit that `E`/`U`/`B` should pass, since a runner holds no contributor public keys. That reasoning is right for GPG. It does not hold for SSH: when `gpg.ssh.allowedSignersFile` is unset, git cannot begin verification at all, so it reports `N` rather than `E`. The previous commit here was SSH signed, which is why it read as unsigned:

```
$ git log --format='%G?' -1 a0430d0                              # locally, allowed_signers configured
G

$ GIT_CONFIG_GLOBAL=/dev/null git log --format='%G?' -1 a0430d0   # what the runner sees
error: gpg.ssh.allowedSignersFile needs to be configured and exist for ssh signature verification
N
```

Two ways to close the gap if you want it closed, both tested locally:

1. Point the workflow at an empty allowed-signers file, which turns `N` into `U` for SSH and the script already passes `U`:
   `git -c gpg.ssh.allowedSignersFile=/tmp/empty log --format='%G?' -1 <sha>` returns `U`.
2. Or test for the signature header directly, which is what the script's comment intends and which covers both formats: `git cat-file commit <sha> | grep -q '^gpgsig'`. That returns 1 for a signed commit and 0 for an unsigned one.

CONTRIBUTING.md does say GPG specifically, so I have moved to GPG here either way and am not asking for the check to change on my account. Happy to send the fix as its own PR if it is wanted, though I did not want to touch #337's work uninvited.

Two practical notes:

- The four workflows on the new head are at `action_required` and need an approval before they re-run.
- Both commits now show as Verified; my GPG public key is registered on my account.

Aman-Mittal (approved):
LGTM thanks for contribution

PR #364, opened 2026-08-16
A11y/icon only button names

opensource-joe:
## What and why

Closes #

## Verification

-

## Checklist

- [ ] I did not hand-edit generated files under `src/app/api/`.
- [ ] New component or service code uses the adapter boundary in `src/app/core/adapters/` instead of direct browser globals or imperative third-party APIs.
- [ ] User-facing strings use translation keys.
- [ ] I added or updated tests appropriate to this change, or explained why tests were not needed.
- [ ] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant.
- [ ] Commits are signed — see [Commit Signing](CONTRIBUTING.md#commit-signing) in CONTRIBUTING.md.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,8 @@ jobs:
         run: npm run i18n:check
       - name: Verify every ion-icon name is registered
         run: npm run check:icons
+      - name: Verify every icon-only button has an accessible name
+        run: npm run check:a11y-names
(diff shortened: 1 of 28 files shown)


opensource-joe:
Closing this as a duplicate. The same work is in #365, which has the fuller description and the verification notes, so please review that one instead.

Apologies for the noise: I opened two PRs for the same change. Identical content in both, 28 files and the same three commits, so nothing is lost here. #365 is the one to look at.

Issue #232, opened 2026-08-02
Working capital, accounting and scheduling screens use the browser's native confirm box (8 sites)


Aman-Mittal:
## What is wrong

Eight destructive actions across **working capital**, **accounting**, **calendars**, **meetings** and **organization** ask for confirmation with `window.confirm` and a hardcoded English string:

```ts
if (!row.id || !window.confirm('Delete this calendar?')) return;
```

That produces the browser's own grey dialog: unstyled, untranslated, unable to signal that the action is destructive, and it freezes the page while it is open. Everywhere else the application uses its own `DialogService.confirm`, which is themed, translated and marks destructive actions in red. Across the app there are 29 of these native calls left; this issue covers the remaining 8.

## Business value

The consequences here are mostly indirect, which is what makes them easy to get wrong. Deleting a provisioning category or criterion changes how loan-loss provisioning is calculated, and the effect shows up later in the accounts rather than immediately on screen. Deleting an account-number format changes how identifiers are generated from that point on, while everything already issued keeps the old shape. Deleting a calendar or meeting removes the schedule a group's collections are organised around.

None of these announce themselves as significant, and the browser's dialog does nothing to help — it shows one English sentence with no context, whatever locale the user picked.

## Files

| File | Line |
|---|---|
| `src/app/features/working-capital/breach/wc-breach-list.component.ts` | 120 |
| `src/app/features/working-capital/loan-products/wc-loan-products-list.component.ts` | 125 |
| `src/app/features/working-capital/near-breach/wc-near-breach-list.component.ts` | 116 |
| `src/app/features/accounting/provisioning-categories/provisioning-categories-list.component.ts` | 114 |
| `src/app/features/accounting/provisioning-criteria/provisioning-criteria-list.component.ts` | 114 |
| `src/app/features/calendars/calendars-list.component.ts` | 128 |
| `src/app/features/meetings/meetings-list.component.ts` | 135 |
| `src/app/features/organization/account-number-formats/account-number-formats-list.component.ts` | 118 |

## How to fix

There is a worked example in the tree — `src/app/features/system/codes/codes-list.component.ts` was converted this way and is the pattern to copy.

```ts
private readonly dialogService = inject(DialogService);
private readonly translate = inject(TranslateService);

void this.dialogService
  .confirm({
    title: this.translate.instant('CALENDARS.DELETE'),
    message: this.translate.instant('CALENDARS.CONFIRM_DELETE', { name: row.title }),
    destructive: true,
[...]

PR #366, opened 2026-08-16
fix(ux): replace native confirm on accounting, organization and working-capital deletes

HurrairaBaloch:
## What and why

Part of #232. Six remaining list screens still confirmed deletes with `window.confirm` — unstyled, English-only, and blocking. They now use `DialogService.confirm` like `codes-list`, with translated copy, `destructive: true`, and the delete skipped when the dialog resolves false.

Calendars and meetings are left for @AmandaKeay as agreed.

Confirm copy is factual (row name only). Provisioning messages do not claim unverified backend consequences; the account-number-format line notes that already issued identifiers keep their current shape.

## Verification

- `npm run lint`
- `npm run i18n:check`
- Prettier on the touched files
- `ng test` for the six list specs: 24/24 passing (including cancel does not call DELETE)
- UI exercised against TestBed mocks, not a live Fineract backend

## Screenshots

Not applicable — same list screens; only the confirm dialog implementation changed.

## Checklist

- [x] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. (unit coverage of confirm/cancel; no e2e for these lists existed)

### DIFF
--- a/src/app/features/accounting/provisioning-categories/provisioning-categories-list.component.ts
+++ b/src/app/features/accounting/provisioning-categories/provisioning-categories-list.component.ts
@@ -19,12 +19,13 @@
 
 import { Component, OnInit, inject, signal } from '@angular/core';
 import { Router } from '@angular/router';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslateModule, TranslateService } from '@ngx-translate/core';
 import { ColumnDef, CellTemplateDirective } from '../../../shared';
 import { DataTableComponent } from '../../../shared/components/data-table/data-table.component';
 import { ProvisioningCategoryService, ProvisioningCategoryData } from '../../../api';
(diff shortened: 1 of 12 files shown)


Aman-Mittal:
superseded by https://github.com/apache/fineract-backoffice-ui/pull/389

Aman-Mittal:
Merged — this is on `main` as b8000280, with you as the author. It went in through the rollup in #389 rather than from this branch, because the commit here was unsigned and the repository requires signatures; cherry-picking and re-signing keeps the authorship and only changes the committer. The change itself is untouched.

Closing this as merged rather than rejected. Thank you — six of the eight sites in #232 are now done. The two left are `calendars-list.component.ts` and `meetings-list.component.ts`, which @AmandaKeay had already said she was starting on, so #232 stays open for those.

PR #382, opened 2026-08-17
fix(a11y): give every products route its own title

opensource-joe:
## What and why

PR 2 of the five agreed on #355. All 65 routes in `products.routes.ts` inherited their tab from the section above them, so every page under `products` read `Products · Fineract`.

This applies the convention settled in #380:

1. The section root restates the `nav.*` key from its `app.routes.ts` entry.
2. Every other route takes the key its page already uses for its own visible heading.
3. Where the heading is a record's name and cannot be a static key, the route takes a generic key naming the screen type.

Refs #355

## The part worth reviewing: this file needed six new keys

`clients` needed none and `system` needs none. `products` needs six, and the reason is a rule about **route shapes** rather than about feature areas, which is the correction I flagged on #380 before it merged.

Five of the six are record views. Rule 2 says take the page's own heading key, and for these that key is `COMMON.DETAILS` or `COMMON.OVERVIEW`, because the visible heading is the record's name and the card above it is generically labelled. Following the rule literally would have given four different screens the tab `Details · Fineract`, reproducing this issue's own problem one level down:

| Route | Title | |
|---|---|---|
| `loan/view/:id` | `nav.loanProductDetails` | existed |
| `savings-accounts/view/:id` | `SAVINGS.ACCOUNT_DETAILS` | new |
| `fixed-deposits/view/:id` | `FIXED_DEPOSITS.ACCOUNT_DETAILS` | new |
| `recurring-deposits/view/:id` | `RECURRING_DEPOSITS.ACCOUNT_DETAILS` | new |
| `shares/view/:id` | `SHARE_ACCOUNTS.ACCOUNT_DETAILS` | new |

The sixth is `ACTIONS.ACCOUNT_ACTION`, for `:accountType/:accountId/action/:command`. That component picks its heading from a command-to-key map at runtime, so no static title can name it. **This is the same shape as the `tellers` `:command` route you asked about on #355**, and `products` has two of them, so the case is now answered in code rather than in prose.

All six are added to `en.json` only. `i18n.js` sets `fallbackLng: "en"`, so the other locales fall back rather than break.

## Verification

| | |
|---|---|
| `npx ng test fineract-backoffice-ui` | **1104 of 1104 SUCCESS**, real headless Chromium |
| `products.routes.spec.ts` alone | 11 of 11 |
| `npm run lint` | exit 0, no new warnings |
| `node scripts/check-translations.mjs` | exit 0, 2206 keys |

**The spec was verified in the failing direction as well.** It carries the two guards from #380 plus one that file did not need:

| Change | Result |
|---|---|
| delete the title on `savings-accounts/view/:id` | 3 failed, naming the route |
| point two view routes at the same `COMMON.DETAILS` | **1 failed, the new uniqueness test** |
| replace a key with the phrase `'Rates'` | 1 failed, the key-shape test |
| restore | 11 of 11 |

That third guard, **no two routes may share a title**, is the one `clients` did not need. Titling 65 routes is only worth doing if the tabs come out distinguishable, and it is what fails if someone later takes the `COMMON.DETAILS` shortcut.

## Screenshots

Not applicable. The change is in the browser tab.

## Two things noticed, neither changed here

**`products/recurring` and `products/share` render the wrong labels.** They are the *product* lists but their headings use `nav.recurringDeposits` and `nav.shares`, which are the *account* labels. Their titles use the correct `nav.recurringDepositProducts` and `nav.shareProducts`, so on those two the tab is right and the on-screen heading is wrong. Worth a separate fix if you agree.

**Three list components carry hardcoded English headings** rather than translation keys: `fixed-deposits-list`, `recurring-deposits-list` and `share-accounts-list` render `title="Fixed Deposit Accounts"` and similar. `check-translations.mjs` cannot catch this, because it only validates strings that are already keys. A real i18n bug but not a titling one, so those three routes take the correct `nav.*` keys and I have left the headings alone. Happy to file it separately.

`products.routes.ts` stays flat, as you asked.

## Checklist

- [x] New component or service code uses the adapter boundary. No new component or service code.
- [x] UI workflow changes include suitable e2e coverage. No workflow changes.
- [x] Commits are signed.

### DIFF
--- a/src/app/features/products/products.routes.spec.ts
+++ b/src/app/features/products/products.routes.spec.ts
@@ -1,0 +1,165 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Component } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { Title } from '@angular/platform-browser';
+import { TitleStrategy, provideRouter } from '@angular/router';
+import { RouterTestingHarness } from '@angular/router/testing';
+import { PRODUCTS_ROUTES } from './products.routes';
+import { TranslatedTitleStrategy } from '../../core/router/translated-title.strategy';
+import { FakeI18nAdapter, provideFakeAdapters } from '../../testing/adapters';
+
+const APP_NAME_KEY = 'app.shortTitle';
+const APP_NAME = 'Fineract';
+const SECTION_KEY = 'nav.products';
+const SECTION_NAME = 'Products';
+
+/** Stands in for the lazily loaded page components, which this spec does not exercise. */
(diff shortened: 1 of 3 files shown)


Aman-Mittal (approved):
LGTM

PR #394, opened 2026-08-19
fix(dialogs): replace native loan and client confirmations

### LINKED ISSUE #231: Loan and client sub-screens use the browser's native confirm box instead of the app dialog (7 sites)
## What is wrong

Seven destructive actions on **loan** and **client** sub-screens ask for confirmation with `window.confirm` and a hardcoded English string:

```ts
if (!row.id || !window.confirm('Delete this charge?')) return;
```

That produces the browser's own grey dialog: unstyled, untranslated, unable to signal that the action is destructive, and it freezes the page while it is open. Everywhere else the application uses its own `DialogService.confirm`, which is themed, translated and marks destructive actions in red. Across the app there are 29 of these native calls left; this issue covers the 7 on loan and client screens.

## Business value

These sit closest to money. Undoing a client transaction, deleting a loan charge, removing a guarantor, dropping an interest pause or a post-dated check all change what a borrower owes or what security stands behind the loan — and every one of them is confirmed today by a browser dialog that a teller can dismiss on reflex without registering what it said.

The transactions one is the sharpest case: the prompt reads "Undo this transaction?" with no amount, no date and no account, so there is nothing on screen to check the action against before agreeing to it. A confirmation that cannot be verified is not really a confirmation.

## Files

| File | Line |
|---|---|
| `src/app/features/loans/charges/loan-charges-list.component.ts` | 115 |
| `src/app/features/loans/guarantors/guarantors-list.component.ts` | 129 |
| `src/app/features/loans/interest-pauses/interest-pauses-list.component.ts` | 112 |
| `src/app/features/loans/post-dated-checks/post-dated-checks-list.component.ts` | 121 |
| `src/app/features/clients/charges/client-charges-list.component.ts` | 121 |
| `src/app/features/clients/collateral/client-collateral-list.component.ts` | 119 |
| `src/app/features/clients/transactions/client-transactions-list.component.ts` | 112 |

## How to fix

There is a worked example in the tree — `src/app/features/system/codes/codes-list.component.ts` was converted this way and is the pattern to copy.

```ts
private readonly dialogService = inject(DialogService);
private readonly translate = inject(TranslateService);

void this.dialogService
  .confirm({
    title: this.translate.instant('LOANS.DELETE_CHARGE'),
    message: this.translate.instant('LOANS.CONFIRM_DELETE_CHARGE', { name: row.name }),
    destructive: true,
  })
  .then((confirmed) => {
    if (!confirmed) return;
    // ...the existing delete call, unchanged
  });
```

Notes:

- `confirm()` returns a `Promise<boolean>`, so the early-return guard becomes a `.then()` block. That is the only structural change.
- Add the new keys to `src/assets/i18n/en.json`. Only `en.json` is required — `npm run i18n:check` fails on keys referenced in code but missing there, and does not require `hi.json`/`ko.json`.
- **Put the identifying detail in the message.** These screens all have the row in hand, so the amount and date can go into the text — the person confirming should be able to check they are undoing the transaction they meant to. Interpolate with `this.translate.instant('KEY', { amount, date })`.
- Set `destructive: true` on every one of these.

## Verifying

```
npm run lint
npm run i18n:check
npm test
npm run build
```

If the screen has a spec, add a case asserting the delete or undo is not called when the dialog resolves `false`. That is the behaviour most likely to break in the rewrite, since the guard moves from a synchronous `return` into a callback — and on these screens a confirmation that silently stops working is a real problem.

## Picking this up

No need to be assigned — assignment here is limited to committers. Comment that you are starting, then open a PR; the comment is enough to stop two people duplicating the work.

This issue touches only files under `src/app/features/loans/` and `src/app/features/clients/`, so it does not overlap with the other `window.confirm` issues and several can be in flight at the same time.

zhangxinyao88:
## What and why

Replace seven `window.confirm` calls on loan and client sub-screens with the app's `DialogService`. The dialogs use translated text, mark the action as destructive, and include values from the selected row so the user can verify the action before confirming.

The six screens with existing specs now cover both confirmation and cancellation.

Closes #231

## Verification

- `npm run lint`
- `npm run i18n:check`
- `npm run format:check`
- `npm run check:icons`
- `npm run check:a11y-names`
- `npm run api:surface`
- `npm run check:licenses`
- `TZ=UTC npm test -- --watch=false` (1155 tests passed)
- `npm run build`

I did not exercise the UI against a live Fineract backend.

## Screenshots

Not included. This change reuses the existing confirmation dialog without changing its appearance.

## Checklist

- [ ] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. No e2e test was added because the change reuses the existing dialog and does not change routing or backend behavior.
- [x] Commits are signed. See [Commit Signing](CONTRIBUTING.md#commit-signing) in CONTRIBUTING.md.

### DIFF
--- a/src/app/features/clients/charges/client-charges-list.component.spec.ts
+++ b/src/app/features/clients/charges/client-charges-list.component.spec.ts
@@ -24,12 +24,14 @@ import { ActivatedRoute, Router, convertToParamMap } from '@angular/router';
 import { of } from 'rxjs';
 import { TranslateModule } from '@ngx-translate/core';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
+import { DialogService } from '../../../core/services/dialog.service';
 
 describe('ClientChargesListComponent', () => {
   let component: ClientChargesListComponent;
   let fixture: ComponentFixture<ClientChargesListComponent>;
   let serviceSpy: jasmine.SpyObj<ClientChargesService>;
   let routerSpy: jasmine.SpyObj<Router>;
+  let dialogService: jasmine.SpyObj<DialogService>;
 
   beforeEach(async () => {
     serviceSpy = jasmine.createSpyObj('ClientChargesService', [
@@ -37,6 +39,8 @@ describe('ClientChargesListComponent', () => {
       'deleteClientsClientIdChargesChargeId',
     ]);
     routerSpy = jasmine.createSpyObj('Router', ['navigate']);
+    dialogService = jasmine.createSpyObj('DialogService', ['confirm']);
+    dialogService.confirm.and.resolveTo(true);
     serviceSpy.getClientsClientIdCharges.and.returnValue(
       of({
         pageItems: [{ id: 1, name: 'Fee', amount: 100 }],
@@ -48,6 +52,7 @@ describe('ClientChargesListComponent', () => {
       providers: [
         { provide: ClientChargesService, useValue: serviceSpy },
         { provide: Router, useValue: routerSpy },
+        { provide: DialogService, useValue: dialogService },
         {
           provide: ActivatedRoute,
           useValue: { snapshot: { paramMap: convertToParamMap({ clientId: '1' }) } },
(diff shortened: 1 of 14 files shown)


Aman-Mittal:
@zhangxinyao88 Pls rebase

zhangxinyao88:
> @zhangxinyao88 Pls rebase

Done

Aman-Mittal (approved):
LGTM and thanks for contribution, Signed commit seems to be flacky so merging this

zhangxinyao88:
I also checked the client transaction undo flow locally after this was merged.

### Before

01-before-current-main-recreated


Before this change, the action used `window.confirm('Undo this transaction?')`.

### After

02-after-shared-dialog

With this change, the dialog shows the transaction ID, amount, date, and explains that the transaction remains marked as reversed.

Behavior check:

- Cancel sent no undo request.
- Confirm sent one undo request.
- After confirmation, the list reloaded and Undo was disabled for the reversed row.

Issue #401, opened 2026-08-19
No repeatable way to populate a backend with demo data, so manual and release-candidate testing starts from an empty tenant


Aman-Mittal:
Part of #395.

## Problem

`e2e/utils/seed-api.ts` has well-tested building blocks for creating offices, clients, loans, savings and so on, but each spec composes only the narrow fixture it needs and tears down around itself. There is no way to say "give me a backend that looks like a working MFI".

Anyone doing manual work against a fresh stack — release-candidate sanity testing (#377), a demo, a screenshot pass, reproducing a bug report — lands on an empty tenant and hand-creates records through the UI before they can test anything. That is slow, and it is inconsistent between people, so "works on a populated tenant" means something different each time.

## Expected

A single command that populates a freshly provisioned Fineract with a representative dataset — an office, staff, a centre, a group, an active loan, a loan pending approval, a savings account, a fixed deposit, a share account, a manual journal entry, reports — and prints what it created.

## Notes

Two constraints matter:

- It must compose the existing `seed-api.ts` helpers rather than introduce a second seeding implementation to keep in step.
- Nothing in it is an assertion, so it must never run as a side effect of `--project=backend` in CI. That argues for its own Playwright project invoked by name, and for not depending on the `setup` project, so it works against a backend that has had nothing else run against it.

Documented in `CONTRIBUTING.md` next to the existing e2e backend instructions.

Issue #400, opened 2026-08-19
RBAC e2e never covers multi-permission routes, ALL_FUNCTIONS_READ, or whether a restricted session survives a reload


Aman-Mittal:
Part of #395.

## Problem

`e2e/rbac-backend-restricted-user.spec.ts` is the only spec driving permissions against a real Fineract. It covers a single-permission route and one action-level gate, which leaves several dimensions of the permission model unproven:

- **Routes declaring more than one permission code.** `permissionGuard` treats multiple codes as OR. Nothing tests that a user holding exactly one of them is admitted, or that a user holding none is refused.
- **`ALL_FUNCTIONS_READ`.** Only ever exercised against a mocked session, never against Fineract's own permission catalogue, so agreement between our shortcut and the backend's is assumed rather than demonstrated.
- **A second action-level gate.** Loan repayment is the only one covered; a single example does not show the pattern holds.
- **Reload persistence.** Restricted permissions are only asserted on a fresh login. Whether they survive an actual page reload — rehydration from storage rather than from the login response — is untested, and that is exactly where a permission set tends to come back wrong.

## Expected

Cover each dimension above, following the existing spec's shape: assert every refusal **both** as "the router sent them to Access Denied" **and** as "Fineract itself returned 403". The guard is defence-in-depth, and showing only the client-side half invites the conclusion that the client-side half is what protects the data.

## Notes

Needs a real backend, so the spec belongs in `BACKEND_SPECS` in `playwright.config.ts`. Some assertions need `data-testid` hooks on action buttons that do not have them yet.

Issue #399, opened 2026-08-19
Clicking a global search result can do nothing — the blur timeout races the click and wins


Aman-Mittal:
Part of #395.

## Problem

`header.component.ts` hides the search results on blur behind a fixed delay:

```ts
onSearchBlur() {
  setTimeout(() => this.showResults.set(false), 150);
}
```

The comment on it is explicit that the delay exists so the list is not unmounted before the click that caused the blur lands. That makes correctness a race against a 150 ms wall clock. On a slow render — a large result set, a throttled CPU, a cold lazy chunk — the timeout fires first, the list collapses out from under the pointer, and the click lands on nothing. The user sees the results vanish and no navigation happen, with no error to explain it.

## Expected

Clicking a result navigates, regardless of render timing. The searchbar should not blur at all when the click target is a result, rather than blurring and hoping the click wins the race.

## Notes

`(mousedown)="$event.preventDefault()"` on the result item stops the blur from being raised for that interaction, leaving the timeout as a fallback for genuine blurs (Escape, clicking outside). The failure is timing-dependent, so a regression test should exercise the collapse path rather than a fast happy path.

Adjacent to #327.

Issue #398, opened 2026-08-19
The dashboard guidance tour highlights the sidebar instead of the System Status card it is describing


Aman-Mittal:
Part of #395.

## Problem

The dashboard tour's second step in `src/app/core/services/guidance.service.ts` declares:

```ts
targetSelector: 'ul',
```

That resolves through `document.querySelector('ul')`, which returns the **first** `<ul>` in the document — the sidebar's own `<ul class="nav-list">`, which renders before the dashboard's main content. The step's copy (`GUIDE.DASHBOARD_ENV_DESC`) describes the "System Overview & Status" card, so a first-time user is told about system status while the tour scrolls to and outlines the navigation menu.

Separately, the tour's own controls in `guidance-tour.component.ts` are hardcoded English:

```html
<ion-button fill="clear" color="medium" (click)="onExit()">Exit</ion-button>
```

`Exit`, `Back`, `Next` and `Finish` all render in English regardless of the selected language — so the onboarding flow, of all things, ignores the language setting. Related to #251 and #388.

## Expected

- The step targets the status card specifically (`.status-list`), not whatever `<ul>` happens to come first.
- Tour controls resolve through `COMMON.*` translation keys.

## Notes

Reproduce by starting the dashboard tour and advancing one step. A regression test needs to assert the highlighted element, not merely that the tour advanced.

PR #402, opened 2026-08-19
feat(working-capital): manage breach, near-breach, delinquency actions and originators

### LINKED ISSUE #396: A Working Capital loan is read-only: delinquency, breach and near-breach actions and originators cannot be recorded
Part of #395.

## Problem

`wc-loan-view.component.ts` shows a Working Capital loan as a Details summary plus read-only tabs. Four write paths that the backend supports have no entry point:

| Capability | Backend service (already generated) | State in UI |
| --- | --- | --- |
| Delinquency actions | `WorkingCapitalLoanDelinquencyActionsService` | Listed, cannot create |
| Breach actions | `WorkingCapitalLoanBreachActionsService` | No tab |
| Near-breach actions | `WorkingCapitalLoanNearBreachActionsService` | No tab |
| Loan originators | `WorkingCapitalLoanOriginatorsService`, `LoanOriginatorsService` | No tab; cannot attach or detach |

Recording a breach action means leaving the back office and calling Fineract directly.

## Expected

- Tabs for breach actions, near-breach actions and originators alongside the existing ones.
- A "New Action" affordance on the delinquency, breach and near-breach tabs, each routing to its own form.
- Originators can be attached to and detached from a loan from the loan itself.

## Notes

Each new route must be guarded on `UPDATE_WORKINGCAPITALLOAN` and pass `scripts/check-route-permissions.mjs`. These endpoints are Fineract 1.15.0+ and unreachable on the public community sandbox, so e2e cover has to mock them.

Aman-Mittal:
Closes #396, #397, #398, #399, #400, #401. Completes the epic in #395.

## Description

A Working Capital loan was read-only in this UI. Four capabilities the backend already exposes — delinquency actions, breach actions, near-breach actions, and per-loan loan-originator attach/detach — had no entry point, so recording a breach action meant leaving the back office and calling Fineract directly.

The main change adds those four write paths. Working through them surfaced several defects and coverage gaps in the same neighbourhood, which are included here rather than deferred; each is filed as its own sub-issue so it can be judged on its own terms.

### Working Capital actions (#396)

- Three new tabs on `wc-loan-view`: breach actions, near-breach actions, originators.
- Three new action-form components, each on its own lazily-loaded route guarded by `UPDATE_WORKINGCAPITALLOAN`, plus a "New Action" affordance on the existing delinquency actions tab.
- Originators can be attached to and detached from a loan directly.
- New `WC_LOANS.*` translation keys; `hi` and `ko` carry the new keys with the same coverage gap as the rest of those files (#255).

### API client sync and the interop call it breaks (#397)

`public/api/fineract.json` re-synced against `apache/fineract:latest` (1472 → 1475 schemas), with provenance updated to the new digest. The sync gives `POST /interoperation/transactions/{accountId}/loanrepayment` a request body where it previously took none — every field optional, so the existing bodyless call still compiled while sending nothing. It now sends `dateFormat` and `locale`, matching every other Fineract command on that screen.

### Dashboard tour highlighted the wrong element (#398)

The tour's second step used `targetSelector: 'ul'`, which resolves to the *first* `<ul>` in the document — the sidebar's `nav-list`, not the System Status card its copy describes. Now targets `.status-list`. The tour's own Exit/Back/Next/Finish buttons were hardcoded English and now resolve through `COMMON.*`.

### Global search swallowed result clicks (#399)

Results were hidden by a 150 ms blur timeout that the click had to beat. Under a slow render the timeout won, the list collapsed, and the click landed on nothing — no navigation, no error. Result items now `preventDefault()` on `mousedown` so the searchbar never blurs for a result click; the timeout remains as a fallback for Escape and click-outside.

### RBAC coverage (#400)

`rbac-multi-permission.spec.ts` covers four dimensions `rbac-backend-restricted-user.spec.ts` does not reach: OR semantics on multi-permission routes, `ALL_FUNCTIONS_READ` against Fineract's real permission catalogue rather than a mocked session, a second action-level gate, and whether a restricted session survives a page reload rather than only a fresh login. Follows the existing spec's shape — every refusal asserted both as "router sent them to Access Denied" and as "Fineract returned 403", since the guard is defence-in-depth and showing only the client half invites the wrong conclusion.

### Demo data seeding (#401)

`npm run seed:demo-data` populates a fresh backend with a representative dataset — office, staff, centre, group, an active loan, a loan pending approval, savings, a fixed deposit, a share account, a journal entry, reports — and prints what it created. Composed entirely from existing `seed-api.ts` helpers, so there is no second seeding implementation to keep in step.

It is its own Playwright project (`demo-seed`) rather than a member of or dependency of `BACKEND_SPECS`: nothing in it asserts anything, so it must never run as a side effect of `--project=backend` in CI. It also does not depend on the `setup` project, so it works against a backend nothing else has touched.

## Checklist

- [x] `npm run lint` clean
- [x] `npm run format:check` clean
- [x] `npm run i18n:check` — no missing keys
- [x] `npm run check:route-permissions` — 309 screens, 119 nav entries, guard order and visibility agree
- [x] `npm run typecheck:e2e` clean
- [x] `npm run check:licenses` — no Category X or unclassified licences
- [x] Unit tests accompany each new component
- [x] Commit is GPG-signed
- [x] Linked to issues in this repository

## Testing notes

The Working Capital endpoints are Fineract 1.15.0+ and not reachable on the public community sandbox, so `working-capital-loan-actions.spec.ts` mocks them. `rbac-multi-permission.spec.ts` needs a real backend and is registered in `BACKEND_SPECS`.


Aman-Mittal:
Superseded by #408.

That branch is this one rebased, with a second commit adding a Vitest runner and a CI ratchet for the deprecated Karma suite. The working-capital commit here is unchanged and carried over intact, so nothing from this PR is lost — it is the first of the two commits on #408, and closes the same issues (#396–#401).

Closing this to keep one branch under review rather than two overlapping ones.

PR #426, opened 2026-08-22
fix(layout): clear the render-time interval when the header is destroyed

### LINKED ISSUE #423: The header's clock timer is never cleared, so every sign-in leaves another one running
Good first issue. One file, a few lines, and the convention to follow is already used elsewhere in this repo.

## What is wrong

`src/app/layout/header.component.ts:465` starts a one-minute interval to refresh the "Render Time" display and never stops it:

```ts
const updateTime = () => {
  this.renderTime.set(
    new Date().toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' }),
  );
};
updateTime();
setInterval(updateTime, 60_000);   // <-- never cleared
```

`HeaderComponent` implements `OnInit` but not `OnDestroy`, and nothing anywhere calls `clearInterval`.

## Why it matters

The header is destroyed and rebuilt whenever the user leaves the authenticated layout — signing out is the ordinary case. The old interval survives that, holding a closure over the destroyed component's `renderTime` signal, and the new header starts another. Sign out and back in three times and three timers are ticking, two of them writing to components nobody can see.

Nothing breaks visibly, which is why it has gone unnoticed — it is wasted work and a small leak that grows with session length. It also keeps firing while the tab is in the background.

This is the **only** `setInterval` in `src/app`, so fixing it here fixes it everywhere:

```
$ grep -rn "setInterval(" src/app --include=*.ts | grep -v spec | grep -v "\.test\."
src/app/layout/header.component.ts:465:    setInterval(updateTime, 60_000);
```

## The fix

Either keep the handle and clear it in `ngOnDestroy`, or — preferred, and already the convention here — let Angular tear it down. `DestroyRef` and `takeUntilDestroyed` are used in `src/app/core/services/idle.service.ts`, `src/app/features/login/login.component.ts` and three other files; follow whichever of those reads closest.

A sketch of the `DestroyRef` form:

```ts
private readonly destroyRef = inject(DestroyRef);
// ...
const id = setInterval(updateTime, 60_000);
this.destroyRef.onDestroy(() => clearInterval(id));
```

## A question worth raising in the PR

While you are in here: "Render Time" shows the viewer's own wall clock, relabelled in developer language, next to "Business Date" — which is the value that actually governs what the platform does. If you think it earns its place in a banking back-office header, say so in the PR; if you think it does not, that is worth a separate issue rather than folding a removal into this fix.

## Verifying

Add a test under `src/app/layout/` (Vitest — `*.test.ts`, see CONTRIBUTING) that destroys the fixture and asserts the timer stopped; `vi.useFakeTimers()` plus `vi.getTimerCount()` is the direct way. Then `npm run test:unit`.

slegarraga:
## What

Closes #423. `HeaderComponent` now implements `OnDestroy`, holds the interval id in a nullable field, and clears it on destroy.

## Why

The header is destroyed and rebuilt whenever the user leaves the authenticated layout — sign-out is the ordinary case. The uncleared interval kept ticking forever, holding a closure over the destroyed component's `renderTime` signal, one more per re-login, still firing in background tabs.

This was the only `setInterval` in `src/app`; with this change none survive their host.

## Verification

- `npm run lint` clean
- `ng build` succeeds
- Unit suite: **431 passed**, with the same single pre-existing failure (`loan-portfolio-summary`) present on main with and without this change — unrelated to this fix

### DIFF
--- a/src/app/layout/header.component.ts
+++ b/src/app/layout/header.component.ts
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-import { Component, inject, signal, OnInit } from '@angular/core';
+import { Component, inject, signal, OnDestroy, OnInit } from '@angular/core';
 
 import { TranslateModule, TranslateService } from '@ngx-translate/core';
 import { AuthService } from '../core/services/auth.service';
@@ -401,7 +401,7 @@ type HeaderSearchResult =
     `,
   ],
 })
-export class HeaderComponent implements OnInit {
+export class HeaderComponent implements OnInit, OnDestroy {
   protected readonly authService = inject(AuthService);
   protected readonly translate = inject(TranslateService);
   protected readonly guidanceService = inject(GuidanceService);
@@ -419,6 +419,7 @@ export class HeaderComponent implements OnInit {
 
   readonly businessDate = signal<string>('-');
   readonly renderTime = signal<string>('-');
+  private renderTimeInterval: ReturnType<typeof setInterval> | null = null;
 
   ngOnInit(): void {
     this.loadSystemInfo();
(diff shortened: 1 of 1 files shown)


Aman-Mittal:
@slegarraga Pls sign your commit and upload your gpg key to your github settings. Your commit show VERIFIED to get merged with mainline

Aman-Mittal (approved):

slegarraga:
Thank you for the review and the merge! Glad the timer leak is closed on main.

PR #429, opened 2026-08-22
test(e2e): await submit enabled before clicking the loan-product form

### LINKED ISSUE #424: e2e: "does not send income recognition on a cumulative product" is flaky under CI-level parallelism
Not blocking anything — it retries green — but it costs a retry cycle on unrelated PRs and trains people to ignore a red run.

## Symptom

`e2e/loan-product-down-payment.spec.ts:327` — *Loan product down payment and tranches › does not send income recognition on a cumulative product* — times out waiting for the submit button:

```
Test timeout of 30000ms exceeded.
Error: locator.click: Test timeout of 30000ms exceeded.
Call log:
  - waiting for getByTestId('loan-product-submit-btn')
    - locator resolved to <ion-button type="submit" ... data-testid="loan-product-submit-btn">
  - attempting click action
    - waiting for element to be visible, enabled and stable

  335 |     await selectOption(page, SCHEDULE_TYPE_LABEL, 'Cumulative');
  336 |
> 337 |     await page.getByTestId('loan-product-submit-btn').click();
```

The locator resolves — the button is on the page. It never becomes *stable and enabled* within 30s.

## It is load-dependent, not code-dependent

Measured on the `mocked` project, same machine, `origin/main` and a feature branch:

| Branch | default workers | `--workers=4` |
|---|---|---|
| `origin/main` | 1/39 failed | 78/78 passed |
| feature branch | 4/39 failed | 78/78 passed |

Both branches fail at default parallelism and both are clean when workers are capped, so this is contention rather than anything a particular change introduced. It predates the branch it was found on.

## Reproducing

```bash
npx playwright test --project=mocked e2e/loan-product-down-payment.spec.ts --repeat-each=3
```

Expect a failure or two in 39 runs on a loaded machine. Capping `--workers=4` hides it, which is the tell.

## Worth checking

The 30s timeout with the element already resolved points at "enabled and stable" rather than at a slow render:

- **Is the button disabled while the form settles?** The test sets several selects in sequence; if validity briefly flips, an `[disabled]` binding could hold the button disabled longer under load. `fixtures.ts` already documents reactive-form validity as the outstanding source of NG0100 in this suite, which is the same territory.
- **Is something still animating?** "Stable" means two consecutive stable bounding boxes; an Ionic overlay closing slowly under CPU contention would defer the click.

Awaiting the button's enabled state explicitly, rather than relying on the click's implicit wait, would both fix it and say what the test is actually waiting for.

Please avoid "fixing" it by raising the timeout — that hides the signal without answering why a click needs 30s.

slegarraga:
## What

Closes #424 — the diagnosis-first way: no timeout was raised; the test now says what it waits for.

## Change

New `submitLoanProduct()` helper: awaits `expect(...).toBeEnabled()` before clicking, then all six submits in the spec use it.

The flake's shape (locator resolved, never *enabled and stable*) pointed at the button sitting `[disabled]` while the reactive form settles. Under default worker parallelism that settle window can stretch past the 30s click timeout; capping workers hid it, which is why CI looked clean at `--workers=4`.

## Verification

| Run | Workers | Result |
|---|---|---|
| This branch | 4 | **39/39 passed** (`--repeat-each=3`) |
| This branch | default | **39/39 passed** |
| This branch, second run | default | **39/39 passed** |
| `origin/main` baseline | default | 39/39 this time — consistent with the issue's "1-4 of 39 on a loaded machine" intermittent profile |

`npm run lint` clean.

### DIFF
--- a/e2e/loan-product-down-payment.spec.ts
+++ b/e2e/loan-product-down-payment.spec.ts
@@ -178,6 +178,19 @@ async function login(page: Page): Promise<Captured> {
   return captured;
 }
 
+/**
+ * Submits the loan-product form. The submit button is disabled while the
+ * reactive form settles; under parallel workers that settle window can stretch,
+ * and a bare click then burns its whole 30s timeout on 'enabled'. Waiting for
+ * the enabled state explicitly says what we are waiting for — and fails fast
+ * with a readable message if the form never becomes valid.
+ */
+async function submitLoanProduct(page: Page): Promise<void> {
+  const submit = page.getByTestId('loan-product-submit-btn');
+  await expect(submit).toBeEnabled();
+  await submit.click();
+}
+
 async function fillRequiredFields(page: Page, name: string): Promise<void> {
   await page.getByRole('textbox', { name: 'Name', exact: true }).fill(name);
   await page.getByRole('textbox', { name: 'Short Name' }).fill('DP01');
@@ -240,7 +253,7 @@ test.describe('Loan product down payment and tranches', () => {
     await page.getByTestId('loan-product-multi-disburse').click();
     await page.getByTestId(TRANCHE_COUNT_TESTID).locator('input').fill('3');
 
-    await page.getByTestId('loan-product-submit-btn').click();
+    await submitLoanProduct(page);
 
     await expect.poll(() => captured.body).not.toBeNull();
     expect(captured.body).toMatchObject({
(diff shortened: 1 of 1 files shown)


Aman-Mittal (approved):
lgtm and thanks for contribution

slegarraga:
Thank you for the review and the merge! Glad the explicit await-toBeEnabled approach landed cleanly, and that both e2e projects stayed green on it.

PR #431, opened 2026-08-22
fix(accounting): add route titles

### LINKED ISSUE #420: Accounting screens show no breadcrumb, and every one of them shares the tab title "Accounting"
Good first issue. Self-contained, one file, no backend needed.

## What is wrong

`src/app/features/accounting/accounting.routes.ts` declares **28 routes and not one `title`**. Two things follow from that, both visible without any setup:

1. **No breadcrumb.** The trail added in #419 is derived from each activated route's `title`. With no title on the child route, only the section contributes a crumb, and the component hides a one-crumb trail — so the whole Accounting section renders with no breadcrumb at all.
2. **Every Accounting screen has the same browser-tab title.** `TranslatedTitleStrategy` falls back to the nearest ancestor that has one. Journal Entries and Chart of Accounts — different screens — both title the tab `Accounting · Fineract`, so browser history and a row of open tabs cannot tell them apart.

## Evidence

Captured against a live Fineract instance.

Journal Entries — no breadcrumb above the content, tab title `Accounting · Fineract`:

Journal entries with no breadcrumb

For contrast, Loan Products, whose route file does set titles — note `Products › Loan Products` above the card:

Loan products with a breadcrumb

## The fix

Add a `title` to each route in `accounting.routes.ts`, pointing at a translation key — the same key the screen's own heading already uses, wherever one exists.

```ts
{
  path: 'journal-entries',
  canActivate: [authGuard, permissionGuard],
  data: { permissions: 'READ_JOURNALENTRY' },
  title: 'ACCOUNTING.JOURNAL_ENTRIES',   // <-- add this
  loadComponent: () => import('./journal-entries-list.component')...
}
```

`src/app/features/products/products.routes.ts` (65 routes, all titled) is the model to copy.

If a key does not exist yet, add it to `src/assets/i18n/en.json` under the section it belongs to. `npm run i18n:check` fails on a key that is referenced but missing, so it will tell you.

## Verifying

```bash
npm start
```

Open an Accounting screen. The breadcrumb should appear above the content and the browser tab should name the screen rather than the section. `npm run lint && npm run i18n:check` should stay green.

## Scope

Only `accounting.routes.ts`. The other untitled route files are separate issues so several people can work in parallel without conflicting — the full list is on #355.

XonkelX:
## Description

Add translation-backed titles to every Accounting child route so breadcrumbs and browser titles identify the active screen instead of stopping at the generic Accounting parent.

The change reuses existing translation keys and adds the one missing English label for editing financial activity mappings.

Closes #420

## Verification

- `npm run lint`
- `npm run build`
- `npm run i18n:check`
- `npm run check:icons`
- `npm run check:a11y-names`
- `TZ=UTC npm run test:unit` — 57 files, 432 tests passed
- Focused Playwright Accounting and breadcrumb checks — 8 passed
- Prettier check on both changed files
- Structural check — all 28 Accounting routes have titles and every key resolves in `en.json`

## Test environment note

The legacy Karma suite was also run on Chrome Headless 151 / Windows: 886 of 898 tests passed. The remaining failures are in unchanged specs and accompanied by the existing Ionic invalid-base-URL harness warnings; a focused Accounting run passed 55 of 56 unchanged component tests. No Karma test imports or exercises the changed route configuration.

## Screenshots

Not included: this is route metadata with no visual redesign. The breadcrumb behavior and Accounting destinations were verified in Chromium through the focused Playwright run.

### DIFF
--- a/src/app/features/accounting/accounting.routes.ts
+++ b/src/app/features/accounting/accounting.routes.ts
@@ -26,6 +26,7 @@ export const ACCOUNTING_ROUTES: Routes = [
     path: 'chart-of-accounts',
     canActivate: [authGuard, permissionGuard],
     data: { permissions: 'READ_GLACCOUNT' },
+    title: 'nav.chartOfAccounts',
     loadComponent: () =>
       import('./chart-of-accounts.component').then((m) => m.ChartOfAccountsComponent),
   },
@@ -33,6 +34,7 @@ export const ACCOUNTING_ROUTES: Routes = [
     path: 'chart-of-accounts/create',
     canActivate: [authGuard, permissionGuard],
     data: { permissions: 'CREATE_GLACCOUNT' },
+    title: 'ACCOUNTING.CREATE_GL_ACCOUNT',
     loadComponent: () =>
       import('./gl-account-form.component').then((m) => m.GLAccountFormComponent),
   },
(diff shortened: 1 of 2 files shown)


Aman-Mittal (approved):

PR #437, opened 2026-08-23
feat(config): customize branding, navigation and labels from a deployment overlay

### LINKED ISSUE #436: Deployment customization without forking: branding, navigation and labels from a config overlay
## Problem

Downstream deployments want their own colours, logo, product name and menu, and want to take
upstream releases as a version bump. Today every one of those changes means editing a file
upstream also edits, so the deployment becomes a fork that conflicts on each release.

The goal is not a better rebase — it is that the common case needs no repository at all:

```dockerfile
FROM apache/fineract-backoffice-ui:1.0.0
COPY branding/ /usr/share/nginx/html/branding/
```

The design test for every customization point: **does upstream ever write this file?** If yes, a
downstream editing it will conflict, however configurable it looks.

Much of the machinery is already here — a runtime `config.json` merged over compiled defaults, a
complete design-token layer in `_common.scss` that Ionic reads through `_ionic-theme.scss`, the
adapter boundary from ADR-0003, and a `NAV_OVERRIDES` token already in front of the navigation
tree. The gap is that these stop one step short of the deployment.

## Defects found while scoping (all silent failures)

**A. `labelKey` is not a stable identity.** `NavOverrides.hidden` matches on `labelKey`, but that
is a field upstream changes freely — the tree already mixes `nav.dashboard`, `SIDEBAR.SEARCH`,
`LOAN_ACCOUNT_LOCK.TITLE` and the raw literal `'Campaigns'`. After a rename, a deployment's
`hidden` entry stops matching and the menu they meant to suppress reappears in production with no
error. Dividers are `{ labelKey: '', divider: true }`, so `"hidden": [""]` removes every divider
at once.

*This one is time-sensitive.* The tree is at `1.0.0-rc.1` with nothing released, so re-keying
`hidden` onto a stable `id` costs nothing today and becomes a breaking change to a documented
config format the day 1.0.0 ships.

**B. The container entrypoint overwrites deployment config.** `deploy/entrypoint.sh` writes
`config.json` whole. Its comment states "every key now has a default here… no key can go missing",
but the heredoc emits five keys and omits `allowedApiOrigins`, `institutionFeatures` and `nav`. A
deployment that bakes any of those into the image loses them at container start —
`allowedApiOrigins` backs a security control, so this is more than cosmetic.

**C. The released image ships a federation host with no remote behind it.** `deploy/Dockerfile`
runs `ng build` on the default project only and never builds or copies `fineract-mfe`;
`src/main.ts` calls `initFederation` unconditionally with `'./remoteEntry.json'`; and
`nginx.conf.template` ends in `try_files $uri $uri/ /index.html`, so that path returns
`index.html` with a `200` rather than a 404. Federation therefore parses HTML as JSON on every
page load. The `.catch(...).then(() => import('./bootstrap'))` chain absorbs it so the app still
boots, but it puts a parse error in every production console and ships a demo scaffold in a
release artifact. Separately, `isLocalDevServer` keys on `location.hostname === 'localhost'`, so
running the released container on localhost requests `http://localhost:4201` — refused by
`connect-src 'self'`.

*(Traced through the build path by inspection; not yet reproduced against a running image.)*

## Scope for 1.0

Filtered on three questions: does it add a dependency (no new npm package **and** no reliance on a
service outside the container), does it widen the deployed CSP or reach an external origin, and
does it freeze a contract we owe compatibility on. Everything below is zero-dependency,
same-origin, and a no-op when the deployment supplies nothing.

- [ ] Stable `id` on every `NavItemConfig`; `hidden` re-keyed onto it; `scripts/check-nav-ids.mjs`
      enforcing presence, uniqueness and stability
- [ ] Layered config resolution with a deployment-owned `branding/config.json` overlay
      (404-tolerant, deep-merged) — fixes B by construction
- [ ] Federation remote map made conditional, so an image with no plugins initialises empty and
      silent — fixes C
- [ ] `BrandingService`: token allow-list, logo, app name, favicon, with a WCAG 4.5:1 contrast
      floor and fallback. Same-origin assets only, so `img-src 'self'` is untouched
- [ ] Navigation composition — `hidden`, per-entry `overrides` (rename / re-icon / reorder /
      re-parent), and deployment-added `items`, all keyed on `id` and passing the same permission
      and feature gates as built-in entries
- [ ] External nav entries that open in a new tab. A top-level link navigation is not a fetch,
      script load or frame, so no CSP change is involved
- [ ] Deployment i18n overlay merged over the shipped catalogue, via the `setTranslation(…, merge)`
      already on the `I18N` adapter contract
- [ ] `/branding/` reserved and gitignored, with CI failing any PR that adds a tracked file there
- [ ] Published JSON Schema so the overlay autocompletes and validates in an editor
- [ ] Reference-downstream fixture in CI: a maximally customized overlay loaded on every PR,
      asserting every id still resolves and every token still applies
- [ ] Startup diagnostics for unknown ids and unrecognised tokens — a typo must not present as a
      missing feature

## Deferred past 1.0, with reasons

| Item | Why it waits |
|---|---|
| Published plugin host API | A host contract in 1.0.0 is a compatibility obligation for the life of the major, and it has had no external consumer to validate it. Wants its own artifact and cadence. |
| Federated remotes read from config | Inert without the API above — a mounted remote could read no session, raise no dialog, share no HTTP chain. Cross-origin also needs `script-src`. |
| Iframe nav entries | Needs `frame-src`. The new-tab form ships in 1.0 and covers the common case. |
| External-origin logos | Needs `img-src`. Mounting under `/branding/` covers it. |
| Branding fetched from a URL | Needs `connect-src` plus a running service. When it lands it must stay optional and provider-agnostic — presentation config must not require a backend. |

Notably, **everything deferred is developer-tier**. No deferred item is something a non-developer
could have used, so the configure-only tier ships complete in 1.0.

## Note

`/admin/wc-cob-tools` is currently cross-listed under both Working Capital and Admin. That looks
deliberate, but it means the same route has two nav entries and appears twice in
`searchRoutes()` / `navDestinationsForPermissions()` results. Worth confirming it is intended.

## Open questions

1. How wide is the published token allow-list? It is the only permanent contract in the 1.0
   scope. Suggest starting with the `:root` block in `_common.scss` and nothing else.
2. Nav ids derived from route, or always explicit? Group headers have no route, so a hybrid means
   two rules. Suggest explicit, enforced by CI.
3. Fix C by making the remote conditional, or by dropping `fineract-mfe` from the release
   entirely? Either is defensible; shipping as-is is not.
4. Do overrides need an in-app editor at 1.0? A schema plus editor autocomplete may be enough, and
   an admin screen writing config back re-opens the backend-dependency question.

Aman-Mittal:
Closes #436 for the 1.0 scope.

## Why

A deployment that wants its own colours, logo, product name or menu currently has to edit files
upstream also edits, so it becomes a fork that conflicts on every release. This gives it a file
upstream never writes instead:

```dockerfile
FROM apache/fineract-backoffice-ui:1.0.0
COPY branding/ /usr/share/nginx/html/branding/
```

Upgrading is then a version bump. No rebase, no merge, no shared file to conflict on.

## What it looks like

The same application, with and without an overlay mounted. Everything below is one JSON file and
two SVGs — no code, no rebuild, no CSP change.

| Stock | With a deployment overlay |
| --- | --- |
|  |  |

Clients and Groups are renamed and pulled to the top, SPM / Working Capital / Interop are gone,
the sidebar takes the institution's navy, and "Logout" reads "Sign out" from the string overlay.

| Dark mode | The sign-in screen |
| --- | --- |
|  |  |

Dark mode uses the lighter accent the overlay names, and the label colour on it is *derived*
rather than fixed — see the contrast note below. The sign-in screen is branded too, since it is
the first thing anyone sees.

The institution's own systems, linked from the menu:

*(Screenshots are generated from the reference overlay against the mocked backend, so they are
reproducible. They live on a branch in a fork rather than in this PR, so no binaries enter the
Apache repository.)*

## How it is layered

| | Layer | Owner |
| - | - | - |
| L0 | `DEFAULT_CONFIG` compiled in | upstream |
| L1 | `/config.json` shipped in the image | upstream |
| L2 | `/config.json` rendered by `deploy/entrypoint.sh` | operator |
| L3 | **`/branding/config.json`** | **the deployment** |
| L4 | the user's stored preferences | end user |

L3 is gitignored, 404-tolerant and deep-merged, and `check-branding-path.mjs` fails the build if
anything is ever committed there — so the promise is enforced rather than documented.

## Three silent failures fixed on the way

**`labelKey` was not an identity.** `NavOverrides.hidden` matched on `labelKey`, a field upstream
renames freely — the tree already mixes `nav.dashboard`, `SIDEBAR.SEARCH`, `LOAN_ACCOUNT_LOCK.TITLE`
and the literal `'Campaigns'`. After a rename the override stopped matching and the menu a
deployment meant to suppress reappeared in production with no error anywhere. Dividers are
`{ labelKey: '' }`, so `"hidden": [""]` removed every divider at once.

Every entry now carries a stable `id`, and `check-nav-ids.mjs` holds them present, unique and
stable. Ids are group-qualified because `/admin/wc-cob-tools` is cross-listed under both Working
Capital and Admin, so a route-derived id would collide for two deliberate entries.

*This is why the PR is worth landing before the tag:* re-keying costs nothing today and becomes a
breaking change to a documented config format the day 1.0.0 ships.

**The entrypoint overwrote deployment config.** `deploy/entrypoint.sh` writes `config.json` whole,
and its comment claimed "every key now has a default here… no key can go missing". The heredoc
emits five keys and omits `allowedApiOrigins`, `institutionFeatures` and `nav`, all added to
`AppConfig` afterwards — so a deployment that baked any of those into the image lost them at
container start, `allowedApiOrigins` included, which backs a security control.

**The released image shipped a federation host with no remote.** `deploy/Dockerfile` builds only
the default project and never copies `fineract-mfe`; `src/main.ts` declared the remote
unconditionally; and `nginx.conf.template` answered the missing `/remoteEntry.json` from
`try_files … /index.html` with a `200`. `initFederation` therefore parsed HTML as JSON on every
page load. The rejection was caught so the app still booted, but every production session opened
with an error in the console. `main.ts` now probes before declaring the remote, and nginx returns a
real 404 for paths that are fetched rather than navigated to.

## What a deployment can set

- **`branding`** — app name, logo, favicon and an allow-listed set of design tokens, applied as CSS
  custom properties.
- **`nav`** — `hidden`, per-entry `overrides` (rename / re-icon / reorder / re-parent) and added
  `items`, all keyed on `id` and passing the same permission and feature gates as built-in entries.
- **`branding/i18n/{lang}.json`** — merged over the shipped catalogue.

`DOCS/CUSTOMIZATION.md` documents it; `public/schema/config.schema.json` gives editors autocomplete
and validation; `DOCS/examples/branding-config.example.json` is a worked example that CI resolves
against the live tree on every PR, and that `npm run branding:demo` mounts for local development.

## Two things worth a reviewer's eye

**`_ionic-theme.scss` referenced the palette by hex, not by token.** Branding would have recoloured
the app chrome and left every Ionic component on the shipped blue. It now reads the tokens, and
`BrandingService` derives the `-rgb`, `-shade`, `-tint` and `-contrast` companions, which CSS cannot.

**The contrast floor applies to two tokens, not all of them.** `secondary-color` and
`primary-strong` must clear 4.5:1 against white, because the stylesheets pair them with a literal
`#fff` and there is no label variable to flip. Every other colour derives its label from the fill —
white and black are the extremes of the WCAG formula and the better of the two bottoms out at
**4.58:1**, so no colour can fail and a floor there would be unreachable code. That is what makes the
lighter dark-mode accent in the screenshot above legal.

## Deliberately out of scope

Mounting a compiled remote from configuration, and the plugin host API it needs. A host contract
published in 1.0.0 is a compatibility obligation for the life of the major version and has had no
external consumer to validate it. `kind: "external"` covers side-by-side systems in the meantime.
Iframe embedding, external asset origins and a remote `brandingUrl` are also deferred: each needs a
CSP directive loosened, and each has a same-origin form shipping here.

Notably, everything deferred is developer-tier — nothing a non-developer could have used.

## Constraints held

- **No new dependency.** Nothing added to `package.json` bar three scripts.
- **The deployed CSP is unchanged.** Assets are same-origin; external nav entries are plain links,
  which the policy does not govern.
- **No backend involved.** The overlay is a static file; nothing here calls Fineract.
- **A no-op when absent**, which is the state every existing deployment is in.

Also migrates off `APP_INITIALIZER` and `provideAnimationsAsync`, both deprecated. Nothing declares
`@angular/animations` triggers and Ionic drives its own transitions.

## Verification

| | |
| - | - |
| Karma | 826 passed |
| Vitest | 527 passed (19 new, `nav-composition.test.ts`) |
| Playwright `mocked` | 26 new across `deployment-customization.spec.ts` and `branded-deployment.spec.ts` |
| lint / format / typecheck (app + e2e) | clean |
| `check:nav-ids`, `check:branding-path`, `check:reference-downstream` | new, wired into CI |
| `i18n:check`, `check:icons`, `check:route-permissions`, `check:version` | pass |

`branded-deployment.spec.ts` renders the published example rather than a fixture, and reading it
from disk is deliberate: a hand-copied overlay drifts, and then the suite passes while the example
deployments actually copy is broken. One artefact, checked three ways — resolved by CI, rendered by
that spec, mounted by the demo script. It paid for itself immediately by catching three defects
that per-mechanism tests had missed: dark-mode branding never applying (inline styles outrank any
stylesheet rule), string overrides losing a race with ngx-translate's own loader, and the sign-in
screen staying unbranded.

Three new JSON files are added to `.rat-excludes` — JSON admits no comments, and all three are data
rather than creative work.

## Not included

The responsive/mobile track from #436. It is independent of this work and shares only the token
allow-list.


Aman-Mittal:
## Colour profiles

Four invented institutions, each one a `branding.tokens` block and a logo — no code, no rebuild,
no CSP change. Note the corner radius moves with them: `--border-radius` is on the allow-list too,
so density and shape travel with the palette, not just hue.

| | Light | Dark |
| --- | --- | --- |
| **Any Community Bank** `#0b5f8a` |  |  |
| **Green Valley Credit Union** `#1b6b3a` |  |  |
| **Meridian Microfinance** `#6b3fa0` |  |  |
| **Sahel Rural Bank** `#9c4221` |  |  |

A worked profile, in full:

```json
{
  "branding": {
    "appName": "Green Valley Credit Union",
    "logoUrl": "branding/logo.svg",
    "tokens": {
      "light": {
        "primary-color": "#1b6b3a",
        "primary-strong": "#14522c",
        "secondary-color": "#123524",
        "border-radius": "10px"
      },
      "dark": { "primary-color": "#5cc98a" }
    }
  },
  "nav": { "hidden": ["spm", "working-capital", "interop"] }
}
```

Each profile names a lighter accent for dark mode, and all eight palettes clear the contrast
floor. The generator asserts `--ion-color-primary` actually equals the deployment's colour before
capturing, so these cannot quietly become screenshots of the shipped blue.

### A correction to the dark-mode screenshot in the description

The earlier dark capture showed a light header. Two things were behind that, and only one was a
real defect:

- **A capture artifact.** The header animates its background over `0.2s` and the screenshot fired
  before the transition settled. Measured after it settles, the header is `rgb(30,30,30)` as
  intended. The image in the description has been regenerated.
- **A real bug, now fixed** (`f852fcd1`). The branding stylesheet is appended after the
  application's own, and `:root` ties with `[data-theme='dark']` on specificity — so any colour a
  deployment set for light and not for dark was also winning dark mode. That is worse than a wrong
  shade: `--secondary-color` is the sidebar ground in light and a near-white *foreground* in dark,
  so the reference deployment's navy was being painted as text on a near-black page.

  Colours the application re-themes are now emitted under `:root:not([data-theme='dark'])`, which
  outranks its `:root` block in light and does not match in dark. Lengths and the colours the
  application does not re-theme stay on `:root` and apply to both. `branded-deployment.spec.ts`
  now asserts the reference overlay's light-only `--secondary-color` does not reach dark mode
  while its `--border-radius` does.

PR #455, opened 2026-08-25
fix(a11y): replace native confirm with DialogService in calendars and meetings

### LINKED ISSUE #232: Working capital, accounting and scheduling screens use the browser's native confirm box (8 sites)

AmandaKeay:
## What and why

Replaces the native `window.confirm()` dialog with the app's styled `DialogService.confirm()` in the calendars and meetings delete flows, matching the pattern already used elsewhere in the app (e.g. `codes-list.component.ts`). This gives these two remaining screens a translated, themed, destructive-marked confirmation instead of the browser's untranslatable, unstyled dialog.

Closes #232

## Verification

- Ran `npm run lint`, `npm run i18n:check`, and `npm test` — all passing
- Updated unit tests in `calendars-list.component.spec.ts` and `meetings-list.component.spec.ts` to mock `DialogService.confirm()` instead of `window.confirm`
- Manually verified the styled confirm modal renders correctly with real translated text (title, message with interpolated name/date, destructive styling) using a temporary preview test with the real `DialogService` and `TranslateModule`
- Exercised with mocks only; did not test against a real Fineract backend

## Screenshots
Before:
Before

After:
After

## Checklist

- [ ] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. — Not added: this is a UI-only confirmation dialog swap with existing unit test coverage; no new user flow to cover with e2e.

### DIFF
--- a/src/app/features/calendars/calendars-list.component.test.ts
+++ b/src/app/features/calendars/calendars-list.component.test.ts
@@ -25,12 +25,14 @@ import { ActivatedRoute, Router, convertToParamMap } from '@angular/router';
 import { of, throwError } from 'rxjs';
 import { provideTranslateTesting } from '../../testing/i18n-testing';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
+import { DialogService } from '../../core/services/dialog.service';
 
 describe('CalendarsListComponent', () => {
   let component: CalendarsListComponent;
   let fixture: ComponentFixture<CalendarsListComponent>;
   let serviceSpy: SpyObj<CalendarService>;
   let routerSpy: SpyObj<Router>;
+  let dialogService: SpyObj<DialogService>;
 
   beforeEach(async () => {
     serviceSpy = createSpyObj([
@@ -38,6 +40,8 @@ describe('CalendarsListComponent', () => {
       'deleteEntityTypeEntityIdCalendarsCalendarId',
     ]);
     routerSpy = createSpyObj(['navigate']);
+    dialogService = createSpyObj(['confirm']);
+    dialogService.confirm.mockResolvedValue(true);
     serviceSpy.getEntityTypeEntityIdCalendars.mockReturnValue(
       of([{ id: 1, title: 'Weekly Meeting', startDate: '2026-01-15' }]) as unknown as ReturnType<
         CalendarService['getEntityTypeEntityIdCalendars']
@@ -50,6 +54,7 @@ describe('CalendarsListComponent', () => {
         ...provideTranslateTesting(),
         { provide: CalendarService, useValue: serviceSpy },
         { provide: Router, useValue: routerSpy },
+        { provide: DialogService, useValue: dialogService },
         {
           provide: ActivatedRoute,
           useValue: {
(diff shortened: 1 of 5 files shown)


Aman-Mittal:
@AmandaKeay please rebase

AmandaKeay:
@Aman-Mittal Rebased onto current main. Both spec files had migrated to Vitest since I originally opened this, so I updated the DialogService mocks to match that style. No more conflicts, ready for review.

Aman-Mittal (approved):

PR #463, opened 2026-08-27
fix(accounting-rules): show load errors instead of an empty list

cestercian:
Part of #223. accounting-rules-list now binds hasError/retry so a failed load is not an empty table.


Aman-Mittal:
@cestercian Can you also share a video or screenshot of your work

PR #495, opened 2026-09-06
fix: register navigation-only icons

### LINKED ISSUE #485: Two navigation icons are never registered with Ionicons, so they silently render as nothing
## What happens

Two icons used by the navigation config fail to load on every page, warning twice per route:

```
[Ionicons Warning]: Could not load icon with name "flash-outline". Ensure that the icon is
registered using addIcons or that the icon SVG data is passed directly to the icon component.

[Ionicons Warning]: Could not load icon with name "checkmark-done-outline". ...
```

Wherever these are used, the `<ion-icon>` renders as empty space — no icon, no fallback, no visible error. It reads as a mildly misaligned label rather than a missing asset.

## Where they come from

Both names are referenced by the navigation config:

```
src/app/core/services/navigation-config.service.ts:638:   icon: 'flash-outline',
src/app/core/services/navigation-config.service.ts:711:   icon: 'checkmark-done-outline',
```

...and `addIcons` appears **nowhere** in `src/app`:

```console
$ grep -rn "addIcons" src/app --include=*.ts
(no matches)
```

So the app relies entirely on Ionicons' lazy CDN-style fetch by name. That is what the warning is complaining about, and it is also why the two failures are silent: the fetch fails and the component renders empty rather than surfacing an error.

This is the same subsystem as the `Failed to construct 'URL': Invalid base URL` exception that fires immediately before each of these warnings — that one is filed separately, but a fix that registers icons explicitly would likely resolve both.

## Suggested fix

Register the icon set explicitly with `addIcons({ flashOutline, checkmarkDoneOutline, ... })` from `ionicons/icons` at bootstrap, rather than depending on name-based lazy loading. That makes the icons resolve deterministically, removes the runtime fetch, and means a typo in a `navigation-config` icon name fails at build time instead of rendering blank in production.

At minimum, the two names above need to resolve to something.

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, Chrome. Both warnings fire on every route load, including a hard reload with cache disabled.

ryux1:
## What and why

Register the `flash-outline` and `checkmark-done-outline` Ionicons referenced by the navigation configuration so those entries no longer render blank or trigger lazy-load warnings.

The existing icon check only inspected `<ion-icon>` template bindings. It now also checks literal and hoisted icon names in the navigation configuration, preventing another navigation-only icon from bypassing CI.

Closes #485

## Verification

- `npm run check:icons` — 117 registered icons
- `npm run lint`
- `npm run format:check`
- `npm run test:scripts` — 28 passed
- `npm test -- --watch=false` — 236 files, 1,382 tests passed
- `npm run build`

## Screenshots

Not captured. This restores the SVGs for two existing navigation entries; the registry check covers both names and the production bundle resolves both imported icon definitions.

## Checklist

- [x] No new component or service code was added; the adapter boundary is unchanged.
- [x] No user-facing strings were added.
- [x] I extended the existing icon registry check to cover navigation configuration.
- [x] No workflow behavior changed; e2e coverage is not applicable to static icon registration.


Aman-Mittal (approved):
lgtm and thanks for comtribution

PR #499, opened 2026-09-06
fix(dashboard): stack widgets on mobile

### LINKED ISSUE #491: The dashboard widget grid never collapses to one column, so at phone widths text wraps one word per line
## What happens

At a standard modern phone width the dashboard's widget area stays in a two-column grid. The columns end up 193px and 123px wide, which is narrower than the content they hold, so headings wrap to two or three lines and body text wraps to one word per line.

Measured at **390×844** on `/dashboard`:

```
.dashboard-layout   grid-template-columns: 193.031px 123px    (total 318px in a 390px viewport)

  .main-column   w 193   heading "Pending Approvals"            → 2 lines
  .side-column   w 123   heading "Loan Status Distribution"     → 3 lines
```

Dashboard at 390x844 — the empty-state sentence wraps to six lines, one word per line, beside a 123px column of stacked headings

In the screenshot, "No items pending approval at this time." wraps to **six lines, one word each**, and the runtime API URL wraps character-by-character. Note also that the grid occupies only 318px of a 390px viewport, leaving a wide empty gutter — the layout is not merely cramped, it is not using the width it has.

## The app already has the pattern this row is missing

The four stat cards directly above this row (Total Clients, Active Loans, Savings Accounts, System Health) collapse to a single column correctly at the same width. So a working single-column mobile treatment already exists in this screen — the widget grid simply was not given the same breakpoint.

That makes this a small, well-scoped fix rather than a redesign.

## Suggested fix

Extend the existing single-column media query to cover `.dashboard-layout`, so the main and side columns stack below the same breakpoint the stat cards already use. Both columns then get the full content width and the wrapping resolves on its own.

Worth checking 360px as well as 390px while fixing, since 360px is still common.

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, backend `sandbox.mifos.community`, Chrome at 390×844 with mobile emulation and touch.

GruffElixir:
## What and why

The dashboard's main and side widget columns remain side by side at phone widths, leaving each too narrow for its headings and status text. The dashboard layout now collapses to one full-width column at the shared 768px mobile breakpoint.

Closes #491

## Validation

- `npm test -- --watch=false --include=src/app/features/dashboard/system-status.component.test.ts` — 7 passed
- ESLint on `system-status.component.ts` — passed

## Checklist

- [x] Existing dashboard unit tests pass.
- [x] Commit is signed.

### DIFF
--- a/src/app/features/dashboard/system-status.component.ts
+++ b/src/app/features/dashboard/system-status.component.ts
@@ -499,6 +499,11 @@ import {
         background: rgba(46, 204, 113, 0.16);
         color: var(--success-color);
       }
+      @media (max-width: 768px) {
+        .dashboard-layout {
+          grid-template-columns: minmax(0, 1fr);
+        }
+      }
     `,
   ],
 })
--- a/e2e/mobile-shell.spec.ts
+++ b/e2e/mobile-shell.spec.ts
@@ -129,6 +129,14 @@ test.describe('the shell at a phone viewport', () => {
     expect(Math.abs(shell - viewport)).toBeLessThanOrEqual(1);
   });
 
+  test('stacks dashboard columns on a phone viewport', async ({ page }) => {
+    const layout = page.locator('.dashboard-layout');
+    await expect(layout).toBeVisible();
+
+    const columns = await layout.evaluate((element) => getComputedStyle(element).gridTemplateColumns);
+    expect(columns.trim().split(/\s+/)).toHaveLength(1);
+  });
+
   describe_drawer();
 
   test('renders tables as cards instead of a sideways scroll', async ({ page }) => {
--- a/e2e/mobile-shell.spec.ts
+++ b/e2e/mobile-shell.spec.ts
@@ -133,7 +133,9 @@ test.describe('the shell at a phone viewport', () => {
     const layout = page.locator('.dashboard-layout');
     await expect(layout).toBeVisible();
 
-    const columns = await layout.evaluate((element) => getComputedStyle(element).gridTemplateColumns);
+    const columns = await layout.evaluate(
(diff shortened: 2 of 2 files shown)


src/app/features/dashboard/system-status.component.ts
```diff
@@ -499,6 +499,11 @@ import {
         background: rgba(46, 204, 113, 0.16);
         color: var(--success-color);
       }
+      @media (max-width: 768px) {
```
Aman-Mittal, line 502: Add screen shot and video and also include dediceted e2e test in mobile view

GruffElixir: Added a dedicated mobile Playwright regression in e2e/mobile-shell.spec.ts. It asserts the dashboard layout resolves to one grid track at the Pixel 7 viewport, covering the mobile stacking behavior directly.

GruffElixir:
The new commit adds the requested mobile regression coverage. GitHub has placed the E2E, CI, and CodeQL runs for commit 4c96c29 in action_required; a repository administrator needs to approve those fork workflows before their results can run.

GruffElixir:
Mobile evidence for the dashboard stacking fix:

- Test: `shell at a phone viewport › stacks dashboard columns on a phone viewport`
- Device profile: Pixel 7, 392×800 viewport
- Result: 1 passed
- The dashboard columns render as a single stacked column at the mobile breakpoint.

dashboard-mobile-5_5
[video.webm](https://github.com/user-attachments/assets/3fd1bab7-8680-4e6e-9311-ccfcd3c19d6d)

Aman-Mittal (approved):

PR #500, opened 2026-09-06
fix(paginator): reflow controls on mobile

### LINKED ISSUE #490: The shared paginator has no mobile layout, so "Items per page:" renders as unreadable fragments on every list screen
## What happens

At phone widths the paginator row tries to fit its label, page-size select, range text and four nav buttons into ~312px on one non-wrapping line. The "Items per page:" label gets squeezed into a 31px-wide box and breaks into fragments that render as stray characters at the left edge — visually just `ıs` and `e:`, with no connection to the words they came from.

Clients list at 368x691 — the paginator shows detached "ıs" and "e:" fragments to the left of the page-size select

Measured at **368×691** on `/clients`:

```
.paginator            width 312   flex-wrap: nowrap   overflow-x: visible

children:
  SPAN.items-per-page   "Items per page:"   w  31   h 54    ← 31px wide, 4 lines tall
  ION-SELECT            page size            w  72   h 44
  SPAN.range-label      "0 of 0"             w  31   h 18    ← nowrap in a 31px box
  ION-BUTTON × 4        nav arrows           w  28   each
```

The label is not merely tight — at 31px for a 15-character string it is genuinely unreadable.

## Cause

`src/app/shared/components/paginator/paginator.component.ts` has no responsive handling:

```css
.paginator         { display: flex; align-items: center; justify-content: flex-end; gap: 8px; ... }
.page-size-select  { min-width: 72px; }
.range-label       { margin: 0 16px; white-space: nowrap; }
```

```console
$ grep -c "@media" src/app/shared/components/paginator/paginator.component.ts
0
```

A fixed `min-width` on the select plus `nowrap` on the range label plus `flex-wrap: nowrap` on the row means the only element that can absorb the shortfall is the text label — so it takes all of it.

## Scope

This is the shared paginator inside `app-data-table`, which is the common list component for the app. **Every list screen is affected on mobile** — Clients, Groups, Loans, Staff, Offices, Users and the rest. Confirmed independently on `/clients` and `/organization/staff` with identical symptoms, which is what establishes it as a shared-component bug rather than a per-screen one.

That also makes it cheap to fix: one component, every screen.

## Suggested fix

Give the paginator an actual mobile layout below a breakpoint rather than a shrunk desktop one:

- drop the "Items per page:" text to just the number + dropdown (the select is self-explanatory next to page arrows), or move it into the select's own label;
- allow the row to wrap, putting the range text and nav arrows on a second line;
- or collapse the page-size selector behind a small menu on narrow viewports.

Removing `white-space: nowrap` alone is not enough — it would wrap the range text instead of fixing the space shortage.

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, Chrome at 368×691 with mobile emulation and touch.

GruffElixir:
## What and why

The shared paginator squeezes its label, page-size selector, range, and four navigation buttons into one non-wrapping row on phones. It now uses a two-row grid at the shared mobile breakpoint: page-size controls occupy the first row, while the range and navigation controls occupy the second.

Closes #490

## Validation

- `npm test -- --watch=false --include=src/app/shared/components/paginator/paginator.component.test.ts` — 12 passed
- ESLint on `paginator.component.ts` — passed

## Checklist

- [x] Existing paginator unit tests pass.
- [x] Commit is signed.

### DIFF
--- a/src/app/shared/components/paginator/paginator.component.ts
+++ b/src/app/shared/components/paginator/paginator.component.ts
@@ -121,9 +121,12 @@ import { PageEvent } from '../../models/table.model';
       }
       @media (max-width: 768px) {
         .paginator {
-          grid-template-columns: auto minmax(0, 1fr) repeat(4, auto);
           display: grid;
+          grid-template-columns: max-content minmax(0, 1fr) repeat(4, auto);
           gap: 4px;
+        }
+        .items-per-page {
+          white-space: nowrap;
         }
         .page-size-select {
           min-width: 0;
--- a/src/app/shared/components/paginator/paginator.component.ts
+++ b/src/app/shared/components/paginator/paginator.component.ts
@@ -119,6 +119,24 @@ import { PageEvent } from '../../models/table.model';
         --padding-end: 6px;
         margin: 0;
       }
+      @media (max-width: 768px) {
+        .paginator {
+          grid-template-columns: auto minmax(0, 1fr) repeat(4, auto);
+          display: grid;
+          gap: 4px;
+        }
+        .page-size-select {
+          min-width: 0;
+        }
+        .range-label {
+          grid-column: 1 / 3;
+          grid-row: 2;
+          margin: 0;
+        }
+        ion-button {
+          grid-row: 2;
+        }
+      }
     `,
   ],
 })
--- a/e2e/mobile-shell.spec.ts
+++ b/e2e/mobile-shell.spec.ts
@@ -161,14 +161,17 @@ test.describe('the shell at a phone viewport', () => {
 
     const paginator = page.locator('app-paginator .paginator');
     await expect(paginator).toBeVisible();
-    const styles = await paginator.evaluate((element) => ({
-      display: getComputedStyle(element).display,
-      columns: getComputedStyle(element).gridTemplateColumns,
-      labelWhiteSpace: getComputedStyle(element.querySelector('.items-per-page')!).whiteSpace,
-    }));
+    const styles = await paginator.evaluate((element) => {
+      const label = element.querySelector<HTMLElement>('.items-per-page')!;
+      return {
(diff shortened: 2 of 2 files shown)


GruffElixir:
@Copilot review

src/app/shared/components/paginator/paginator.component.ts
```diff
@@ -119,6 +119,27 @@ import { PageEvent } from '../../models/table.model';
         --padding-end: 6px;
         margin: 0;
       }
+      @media (max-width: 768px) {
+        .paginator {
+          display: grid;
+          grid-template-columns: max-content minmax(0, 1fr) repeat(4, auto);
```
Aman-Mittal, line 125: Also add e2e coverage for mobile view

GruffElixir:
@Copilot review

GruffElixir:
Addressed the review feedback in e8a5c78: the mobile grid now gives the translated label an intrinsic max-content track, keeps it on one line, and leaves the select as the flexible track. The focused paginator suite passes (12 tests), with formatting, lint, and diff checks clean. @Copilot review

src/app/shared/components/paginator/paginator.component.ts
GruffElixir: Added a dedicated mobile Playwright regression in e2e/mobile-shell.spec.ts. It asserts the paginator uses an intrinsic max-content label track and keeps the translated label on one line at the Pixel 7 viewport.

GruffElixir:
Rebased onto upstream main and resolved the e2e/mobile-shell.spec.ts conflict while retaining both dashboard column-stacking and paginator-label coverage. Pushed d7f9a5f; git diff --check passes. @Copilot review

Aman-Mittal (approved):

PR #502, opened 2026-09-06
fix(login): keep form reachable in short viewports

### LINKED ISSUE #482: The Sign In button falls below a short viewport and the login page cannot be scrolled to reach it
## What happens

On a viewport shorter than roughly 620 CSS pixels, the login card's **Sign In** button renders below the fold — and the page cannot be scrolled, so the button is genuinely unreachable with a pointer. The only way in is to press <kbd>Enter</kbd> from a field, which a first-time user has no reason to try.

Measured at **1280×560**:

```
submit button   top: 578   bottom: 617      (viewport height: 560)
documentElement scrollHeight: 560   clientHeight: 560   →  not scrollable
body            scrollHeight: 706                        →  146px of content clipped away
```

Login card at 1280x560 — the password field is cut off and Sign In is entirely off-screen, with no scrollbar

## Why it cannot be scrolled

`.login-page` in `src/app/features/login/login.component.ts` is `min-height: 100dvh` with the card vertically centred, so when the card is taller than the viewport it overflows top **and** bottom rather than growing the page.

The overflow is then clipped rather than scrolled, because Ionic's global stylesheet sets:

```css
body { overflow: hidden; }
```

Confirmed by walking the live stylesheets:

```js
// rules matching html/body that set overflow
[{ sel: "body.backdrop-no-scroll", overflow: "hidden", href: ".../styles.css" },
 { sel: "body",                    overflow: "hidden", href: ".../styles.css" }]
```

The login screen is a plain `<div class="login-page">` rather than an `<ion-content>`, so it never gets the scroll container that Ionic's `overflow: hidden` on `body` assumes is there. Every other screen in the app sits inside the app shell, which provides its own scrolling area — login is the one route that does not.

## Who this affects

- 1366×768 laptops with a browser toolbar and bookmarks bar (usable height often ~600px)
- Any window the user has not maximised
- Phones in landscape

It is not a niche viewport — 1280×560 is an ordinary maximised window on a common laptop once browser chrome is subtracted.

## Suggested fix

Give the login route its own scroll container instead of relying on the body, e.g. make `.login-page` `min-height: 100dvh; overflow-y: auto;` and switch the centring from a hard `align-items: center` to one that degrades to top-aligned when content exceeds the viewport (`justify-content: safe center`, or `margin: auto` on the card). Tightening the card's vertical padding helps but does not fix the unreachable case on its own.

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, backend `sandbox.mifos.community`, Chrome.

GruffElixir:
## What and why

The login page is taller than short viewports, but Ionic keeps the document body from scrolling. The page now owns a viewport-height scroll container, includes its padding in that height, and uses safe vertical centering so oversized content aligns to a reachable scroll origin.

Closes #482

## Validation

- `npm test -- --watch=false --include=src/app/features/login/login.component.test.ts` — 6 passed
- ESLint on `login.component.ts` — passed

## Checklist

- [x] Existing login unit tests pass.
- [x] Commit is signed.

### DIFF
--- a/src/app/features/login/login.component.ts
+++ b/src/app/features/login/login.component.ts
@@ -183,10 +183,13 @@ import { HelpIconComponent } from '../../shared/components/help-icon/help-icon.c
       .login-page {
         display: flex;
         justify-content: center;
-        align-items: center;
+        align-items: safe center;
         /* dvh: on a phone, 100vh is measured against the retracted URL bar, so the card
            ends up centred against a viewport taller than the screen. */
+        box-sizing: border-box;
+        height: 100dvh;
         min-height: 100dvh;
+        overflow-y: auto;
         background: linear-gradient(135deg, #2c3e50 0%, #3498db 100%);
         padding: 1rem;
       }
--- a/src/app/features/login/login.component.ts
+++ b/src/app/features/login/login.component.ts
@@ -187,6 +187,8 @@ import { HelpIconComponent } from '../../shared/components/help-icon/help-icon.c
         /* dvh: on a phone, 100vh is measured against the retracted URL bar, so the card
            ends up centred against a viewport taller than the screen. */
         box-sizing: border-box;
+        height: 100svh;
+        min-height: 100svh;
         height: 100dvh;
         min-height: 100dvh;
         overflow-y: auto;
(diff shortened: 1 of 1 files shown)


GruffElixir:
@Copilot review

GruffElixir:
@Copilot review

GruffElixir:
Addressed the review feedback in 15e3991: the login page now has a small-viewport height fallback and a plain center alignment fallback before safe center, while retaining the scrollable dvh layout. The focused login suite passes (6 tests), with formatting, lint, and diff checks clean. @Copilot review

Aman-Mittal (approved):

PR #507, opened 2026-09-06
fix(login): stop announcing pristine inputs as aria-invalid

### LINKED ISSUE #483: Login announces Username and Password as invalid before anyone has typed in them
## What happens

On a fresh load of `/login`, before any interaction at all, the Username and Password inputs already carry `aria-invalid="true"`. A screen reader announces both fields as invalid on a form nobody has touched yet.

Measured in a **fresh isolated browser context**, immediately after load, with no clicks or keystrokes:

```json
{
  "tenantId": { "aria-invalid": "false", "value": "default" },
  "username": { "aria-invalid": "true",  "value": "" },
  "password": { "aria-invalid": "true",  "value": "" }
}
```

The accessibility tree agrees:

```
textbox "Username" invalid="true"
textbox "Password" invalid="true"
```

## Cause

`src/app/features/login/login.component.ts` binds the attribute straight to the control's validity, with no check on whether the user has interacted with it:

```
132:  [attr.aria-invalid]="loginForm.get('tenantId')?.invalid"
143:  [attr.aria-invalid]="loginForm.get('username')?.invalid"
154:  [attr.aria-invalid]="loginForm.get('password')?.invalid"
```

`tenantId` reads `false` only because it ships with a default value of `default` — it is not validated differently, it simply happens to be non-empty. That is what confirms the diagnosis: the binding reports "empty required field" as "invalid", which is true of the model but wrong as a user-facing claim about a pristine form.

Note there is no visible red styling, so this is purely an assistive-technology problem: a sighted user sees a normal empty form while a screen-reader user is told two fields are in an error state.

## Suggested fix

Gate the attribute on the control having been interacted with, and emit `null` rather than `false` so the attribute is absent instead of explicitly negative:

```html
[attr.aria-invalid]="
  (c.invalid && (c.touched || c.dirty)) ? 'true' : null
"
```

Worth applying to all three fields together, since `tenantId` has the same binding and would show the same behaviour the moment its default is cleared.

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, Chrome, fresh isolated browsing context.

v01dst:
## Summary

On a fresh `/login` load the username and password inputs carried `aria-invalid="true"` before any interaction, so screen readers announced both fields as invalid on a pristine form (#483). `tenantId` escaped only because it ships with a default value.

## Fix

Each binding now gates on touched/dirty per the issue's suggestion, and emits `null` (attribute absent) rather than `false` for pristine fields:

```
[attr.aria-invalid]="(control.invalid && (control.touched || control.dirty)) ? 'true' : null"
```

After a submit attempt (touched/dirty), genuinely invalid fields still announce `aria-invalid="true"` — the error state is preserved where it is real.

Closes #483

### DIFF
--- a/src/app/features/login/login.component.ts
+++ b/src/app/features/login/login.component.ts
@@ -129,7 +129,7 @@ import { HelpIconComponent } from '../../shared/components/help-icon/help-icon.c
                 id="tenantId"
                 type="text"
                 formControlName="tenantId"
-                [attr.aria-invalid]="loginForm.get('tenantId')?.invalid"
+                [attr.aria-invalid]="(loginForm.get('tenantId')?.invalid && (loginForm.get('tenantId')?.touched || loginForm.get('tenantId')?.dirty)) ? 'true' : null"
               />
             </div>
 
@@ -140,7 +140,7 @@ import { HelpIconComponent } from '../../shared/components/help-icon/help-icon.c
                 type="text"
                 formControlName="username"
                 autocomplete="username"
-                [attr.aria-invalid]="loginForm.get('username')?.invalid"
+                [attr.aria-invalid]="(loginForm.get('username')?.invalid && (loginForm.get('username')?.touched || loginForm.get('username')?.dirty)) ? 'true' : null"
               />
             </div>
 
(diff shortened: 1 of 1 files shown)


v01dst:
Closing this. I was batch-generating PRs across repos while focused on my own project — that wasn't fair on reviewers' time, and I've stopped. Apologies for the noise.

Issue #513, opened 2026-09-06
The interest-pause edit form discards dates the user picked while its own load was still in flight


Aman-Mittal:
## What happens

`interest-pause-form.component.ts` renders its pickers immediately and fetches the pause afterwards. When the response arrives it writes into the same signals the user has been editing:

```ts
// :184
private loadPause(): void {
  if (!this.loanId || !this.variationId) return;

  this.pauseService.getLoansLoanIdInterestPauses(this.loanId).subscribe({
    next: (pauses) => {
      const pause = pauses.find(({ id }) => id === this.variationId);
      if (pause) {
        this.startDate.set(this.toFormDate(pause.startDate));   // unconditional
        this.endDate.set(this.toFormDate(pause.endDate));
      }
    },
```

Nothing disables the form while that request is outstanding and nothing checks whether the user has already touched the fields, so on a slow link the picked dates are replaced by the stored ones.

## Reproduced against a live platform

Loan 26, pause id 1 (2026-08-17 to 2026-08-27), with the populating `GET` held for five seconds — a delay, not a fabricated response. Dates picked while it was in flight:

![Oct 10 and Oct 20, as picked](https://raw.githubusercontent.com/Aman-Mittal/fineract-backoffice-ui/assets/issue-screenshots/interest-pause-overwrite-picked.png)

The same screen once the response landed:

![Back to Aug 17 and Aug 27](https://raw.githubusercontent.com/Aman-Mittal/fineract-backoffice-ui/assets/issue-screenshots/interest-pause-overwrite-reverted.png)

Read off the controls directly:

```
stored pause     : 2026-08-17 .. 2026-08-27
user picked      : {"start":"2026-10-10","end":"2026-10-20"}
after GET landed : {"start":"2026-08-17","end":"2026-08-27"}
```

[Screen recording](https://github.com/Aman-Mittal/fineract-backoffice-ui/raw/assets/issue-screenshots/interest-pause-overwrite.webm).

## Why it matters

The reversion is silent and it lands on a form the user is about to submit. Someone correcting a pause on a slow connection picks the right dates, watches them change back, and — if they do not happen to be looking at that moment — saves the values they meant to replace. The result is an unchanged relief period that the operator believes they corrected, on a loan where the pause range determines interest.

Five seconds is generous for a lab, but a field branch on a poor link is exactly where this bites, and it is also the setting where a user is most likely to start typing before the screen has settled.

## Suggested fix

[...]

PR #523, opened 2026-09-06
fix(config): only fetch the branding overlay when a deployment ships one

### LINKED ISSUE #487: The optional branding overlay is fetched on every load and its expected absence surfaces as two 404 errors
## What happens

Two requests 404 on every page load of a default deployment:

```
GET /branding/config.json?cb=1788679060016   404
GET /branding/i18n/en.json                   404
```

These come from the L3/L4 deployment overlay described in `DOCS/CUSTOMIZATION.md`.

## The files being missing is correct — surfacing it as an error is not

`src/app/core/services/config.service.ts` documents the overlay as expected to be absent:

```
/**
 * ...
 * Gitignored, never written by upstream, and expected to be absent.
 */
const DEPLOYMENT_OVERLAY_URL = 'branding/config.json';
```

The same holds for `OVERLAY_PREFIX = 'branding/i18n/'` in `src/app/core/adapters/i18n/deployment-translate.loader.ts`.

So an install where no deployer has configured white-label overrides — the default, and the state of every fresh deployment — is a fully supported configuration that nonetheless reports two network-level errors on every route. The service's own comment says these files should not be there; the browser says something is wrong. Both cannot be right.

## Why it matters

This app already logs `Failed to construct 'URL': Invalid base URL` twice and two Ionicons warnings per page load (both filed separately). Adding these two 404s means **six recurring console entries on every load before the user has done anything**. A developer debugging a real problem has to learn to filter six known-noisy entries first, which is exactly how a genuine new error gets missed.

It also gives a deployer evaluating the customisation system a false signal that they have misconfigured something when they have not configured anything.

## Suggested fix

Treat a 404 on the overlay as the documented "no overlay configured" case rather than as a failure:

- Probe with a request whose failure is expected and handled, and log at `debug`/`info` — not as an uncaught network error — when it comes back 404.
- Or gate the fetch behind an explicit opt-in (a `brandingOverlayEnabled` flag in `config.json`, defaulting off), so a deployment that has not configured branding never issues the request at all. This also drops two requests from every cold load.

The second is cleaner: the current code cannot distinguish "no overlay configured" from "overlay configured but misdeployed", and a flag makes that difference explicit and reportable.

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, Chrome. Both 404s appear in the network log on every route load.

Aman-Mittal:
Closes #487.

## What it does

`branding/config.json` and `branding/i18n/{lang}.json` were requested on every load. They are absent on every default install — the documented, correct state — so a fresh deployment reported two 404s in the browser console on every route.

The application already declined to *report* those failures: both fetches carry `skipErrorToast` and both resolve to `{}`. That is as far as this side can reach. A 404 is a network-level event the browser writes regardless, so the only way for a supported configuration to stop looking like a misconfiguration is not to make the request.

## The flag, and where it has to live

`brandingOverlayEnabled` gates both fetches. It cannot live in the overlay itself — a file cannot announce its own absence — so it belongs in the layer below.

**The container entrypoint sets it for you**, by looking for the directory the deployer already declared with the `COPY branding/ ...` line in `DOCS/CUSTOMIZATION.md`. There is nothing new to remember, and it satisfies that heredoc's own stated rule:

> the rule for this heredoc is narrow: it carries only what the *container environment* is the right source for

Whether this image carries a branding overlay is a fact about the container's filesystem, which is exactly that. A deployment serving `dist/` from its own web server sets it in `config.json` alongside `fineractApiUrl`; that case is documented.

The flag also makes a distinction the old code could not: "no overlay configured" versus "overlay configured but misdeployed". The second is a real error a deployer wants to see, and it still surfaces — a declared overlay that 404s is tolerated in the merge but is no longer indistinguishable from the ordinary case.

## Verified against a running app

Console across `/dashboard`, `/clients`, `/loans`, `/notifications`, capturing every line.

Before, on `main`:

```
failed net::ERR_CONNECTION_REFUSED  http://localhost:4201/remoteEntry.json
http 404  /branding/config.json
http 404  /branding/i18n/en.json
```

After, on this branch:

```
failed net::ERR_CONNECTION_REFUSED  http://localhost:4201/remoteEntry.json
```

What remains is the microfrontend remote, which is simply not running in a plain `ng serve`.

And the positive path, with the flag on and a real overlay in place:

```
branding requests: 200 /branding/config.json?cb=... | 200 /branding/i18n/en.json
brand text:        Any Community Bank
```

The overlay's `appName` renders and the language overlay is merged, so the flag gates the probe without disabling the feature. The temporary overlay used for that check was removed; `npm run check:branding-path` passes.

## One trap worth flagging in review

The loader resolves `ConfigService` **on use**, not as a field. Injecting it at construction closes a cycle — `ConfigService` needs `HttpClient`, whose `errorInterceptor` needs `I18N`, which needs the loader. Angular reports it as NG0200 during bootstrap and the app renders nothing at all.

The unit tests do not catch that, because they provide `ConfigService` directly. I found it by loading the application; `login.spec.ts` would have caught it in CI. Worth knowing before anyone tidies that `inject(Injector)` back into a field — the comment on it says so.

## Verification

| | |
|---|---|
| `npm run test:unit` | 237 files / 1427 tests, exit 0 |
| `TZ=America/New_York npm run test:unit` | 237 files / 1427 tests, exit 0 |
| `npm run build` | clean |
| `npm run lint` | clean |
| `bash scripts/check-license.sh` | clean |
| `node scripts/check-branding-path.mjs` | `public/branding/` is reserved and empty |
| `bash -n deploy/entrypoint.sh` | clean |

New coverage: four cases on `ConfigService` for the probe (not declared, declared off, declared on, declared but missing) and a new `deployment-translate.loader.test.ts` with four more, including that the overlay request is not issued when undeclared.

Rebased on `f18d9592`.

### DIFF
--- a/DOCS/CUSTOMIZATION.md
+++ b/DOCS/CUSTOMIZATION.md
@@ -52,6 +52,32 @@ start, so anything you put there is erased on the next restart.
 Layers are deep-merged, so naming one key does not discard the rest. Arrays are replaced, not
 concatenated — writing `nav.hidden` means "hide exactly these".
 
+### L3 is only read when the deployment says it exists
+
+`brandingOverlayEnabled` gates whether the app asks for `branding/config.json` and
+`branding/i18n/` at all. It defaults to `false`, and **the container entrypoint sets it for you**
+by looking for the directory — so the `COPY branding/` above is the whole declaration and there
+is nothing extra to remember.
+
+It is the one setting that cannot live in L3: a file cannot announce its own absence.
+
+The flag exists because an install with no overlay is the normal, supported case, and probing for
+it anyway put two 404s in the browser console on every page load. The application can decline to
+_report_ a 404, and does — but it cannot stop the browser writing it. A supported configuration
+should not look like a broken one.
+
+If you serve `dist/` from your own web server rather than the container image, set it yourself
+alongside `fineractApiUrl`:
+
+```json
+{
+  "fineractApiUrl": "/api/v1",
+  "brandingOverlayEnabled": true
+}
+```
+
+Leave it off and `branding/` is never read, whatever it contains.
+
 ## Getting started
 
 Point your editor at the schema and it will autocomplete and validate as you type:
--- a/deploy/entrypoint.sh
+++ b/deploy/entrypoint.sh
@@ -106,16 +106,30 @@ for pair in "RBAC_ENABLED:$FINERACT_UI_RBAC" "DEVELOPER_TOOLS_ENABLED:$FINERACT_
   fi
 done
 
+# Whether this image carries a branding overlay, which is a fact about the container's own
+# filesystem and so exactly what this heredoc is for — the deployer already declared it by
+# doing the `COPY branding/ ...` in DOCS/CUSTOMIZATION.md, and should not have to say it twice.
+#
+# The app cannot work this out for itself. Asking for a file that is absent on every default
+# install put a 404 in the browser console on every page load, and a 404 is a network-level
+# event the application can decline to report but cannot suppress.
+if [ -d "$HTML_ROOT/branding" ]; then
+  FINERACT_UI_BRANDING=true
+else
+  FINERACT_UI_BRANDING=false
+fi
+
 cat > "$HTML_ROOT/config.json" <<EOF
 {
   "fineractApiUrl": "/api/v1",
   "defaultTenant": "$FINERACT_UI_TENANT",
   "rbacEnabled": $FINERACT_UI_RBAC,
   "institutionType": "$FINERACT_UI_INSTITUTION",
-  "developerToolsEnabled": $FINERACT_UI_DEVTOOLS
+  "developerToolsEnabled": $FINERACT_UI_DEVTOOLS,
+  "brandingOverlayEnabled": $FINERACT_UI_BRANDING
 }
 EOF
 
-echo "entrypoint: tenant ${FINERACT_UI_TENANT}, rbacEnabled ${FINERACT_UI_RBAC}, developerTools ${FINERACT_UI_DEVTOOLS}"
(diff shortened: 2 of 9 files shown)


Aman-Mittal:
Pushed `b438c29e` for the 26 mocked-e2e failures. They were mine, and they were pointing at something real about the fixtures.

Both branding suites served `branding/config.json` while the layer beneath it said nothing. That is a state the container entrypoint cannot produce — it sets `brandingOverlayEnabled` from the directory the deployer copied in — so once the probe is gated on that flag, the fixtures were modelling a deployment that shipped `branding/` and never told the application. They now declare it, the way a real deployment does:

- `deployment-customization` derives it from whether the test asked for an overlay at all, so the "no overlay mounted" cases keep exercising the off path.
- `branded-deployment` always mounts one, so it is always on.

Added the assertion this change is actually about, which only an e2e can make:

```ts
test('does not ask for an overlay this deployment has not declared', async ({ page }) => {
  const asked: string[] = [];
  page.on('request', (r) => { if (r.url().includes('/branding/')) asked.push(r.url()); });
  await deployWith(page, null);
  await expect(page.locator('.app-title')).toBeVisible();
  expect(asked).toEqual([]);
});
```

A unit test cannot see this. It is about a request not being on the wire, and the console entry it used to produce was written by the browser, not by us.

```
npx playwright test --project=mocked e2e/deployment-customization.spec.ts e2e/branded-deployment.spec.ts
  28 passed
```

### One local-only failure, for the record

A full `--project=mocked` run here came back `1 failed / 342 passed` on `loan-servicing-gaps.spec.ts:203` — the interest-pause edit, whose Save button stayed disabled because the form had not finished loading its dates.

It passes on its own, and it is a local artifact: this config uses `workers: 1` and `retries: 2` under CI but unbounded workers and no retries locally, so my run was far more contended than CI ever is. Worth noting that the screen it happens on is the one #513 is about — a form that races its own populating request — so the test is inherently timing-sensitive until that is fixed.

Not touched here.

PR #524, opened 2026-09-06
fix(loans): send an unusable interest-pause id back to the list

### LINKED ISSUE #512: A non-numeric :variationId makes the interest-pause screen say Edit and save as Create
## What happens

`/loans/{loanId}/interest-pauses/edit/:variationId` takes its id straight from the route:

```ts
// src/app/features/loans/interest-pauses/interest-pause-form.component.ts:176
if (variationId) {
  this.variationId = +variationId;
  this.isEditMode.set(true);
  this.loadPause();
}
```

The route has no constraint on the segment, so `edit/abc` gives `+'abc'` → `NaN`. From there two guards disagree with `isEditMode`, because `NaN` is falsy:

```ts
private loadPause(): void {
  if (!this.loanId || !this.variationId) return;   // NaN → returns, form stays empty
```

```ts
const save$ =
  this.isEditMode() && this.variationId              // NaN → falsy
    ? this.pauseService.putLoansLoanIdInterestPausesVariationId(...)
    : this.pauseService.postLoansLoanIdInterestPauses(this.loanId, request);  // ← taken
```

So the header reads **Edit Interest Pause**, the pickers sit on today rather than on the stored pause, and pressing Save creates a second pause instead of correcting the first. Nothing on screen says so.

## Reproduced against a live platform

Loan 26 on the local stack: progressive, interest-recalculating, one pause from 2026-08-17 to 2026-08-27.

Navigating to `/loans/26/interest-pauses/edit/abc`:

The edit screen with a NaN id

Picking 1–5 September and saving produced exactly one request, and it is a create:

```
POST /fineract-provider/api/v1/loans/26/interest-pauses
{"startDate":"01 September 2026","endDate":"05 September 2026","dateFormat":"dd MMMM yyyy","locale":"en"}
```

No `PUT .../interest-pauses/{id}`, and no `GET` beforehand — `loadPause()` had already returned. The list afterwards:

Two pauses where the user expected one edit

[Screen recording](https://github.com/Aman-Mittal/fineract-backoffice-ui/raw/assets/issue-screenshots/interest-pause-edit-nan.webm).

## Why it matters

An interest pause changes what a borrower owes. A user who reaches this screen from a stale link, a typo, or a bookmark that outlived a data reset gets a screen that claims to be editing and silently grants a *second* relief period on top of the first. The loan is then paused across two ranges the operator never agreed to, and on a recalculating loan that is a real schedule change, not a display artifact.

The same shape applies to `0` and to any id the segment cannot parse.

## Suggested fix

Reject an unusable id rather than falling through to create. Parse once, and let the rest of the component depend on that one decision:

```ts
const parsed = Number(variationId);
if (!Number.isInteger(parsed) || parsed <= 0) {
  // Not an id this screen can edit — send the user back to the list.
  void this.router.navigate(['/loans', this.loanId, 'interest-pauses']);
  return;
}
```

Whatever the handling, `isEditMode()` and the `save$` branch must not be able to disagree — the bug is that they are derived from the same value by two different truthiness rules. Deriving the mode from `variationId !== null` after validation removes the possibility.

A route matcher constraining `:variationId` to digits would stop it earlier, and is worth doing as well as, not instead of, the guard.

## Regression guard

A unit case asserting that a non-numeric route parameter does not call `postLoansLoanIdInterestPauses`. The existing spec already stubs `LoanInterestPauseService`, so this is one more case in the file, not new scaffolding.

## Environment

`main` at `3c6d7479`. `apache/fineract:latest` via `deploy/docker-compose-e2e.yml`, driven through the UI with no API seeding of the request itself.

Found while reviewing #283 / PR #481 after it merged.

arpitjain099:
The route puts no constraint on `:variationId`, so `edit/abc` gave `+'abc'` and left `variationId` as `NaN`. `isEditMode` was set from the presence of the segment, but `loadPause` and the save branch both test the id itself, and `NaN` is falsy. The screen therefore said Edit Interest Pause, the pickers sat on today, and Save posted a create. Nothing on screen said so, which is the part that makes it worth fixing rather than leaving to the API.

`toRouteId` now parses the segment and returns null for anything that cannot be a pause id: non-numeric, empty, zero, negative, fractional. An id that fails that check sends the user back to `/loans/{loanId}/interest-pauses`, the same place Cancel and a successful save go. That felt more honest than quietly rendering the create form under an `/edit/` URL. If you would rather show an error and stay put, say so and I will change it.

The same helper reads `loanId`, so a non-numeric loan id no longer produces `NaN` there either.

Five parameterized cases in the existing spec cover `abc`, `0`, `-1`, `1.5` and a blank segment, asserting edit mode stays off, no pause fetch is issued and the navigation happens. Against the current component all five fail:

    src/app/features/loans/interest-pauses/interest-pause-form.component.test.ts (10 tests | 5 failed)

and they pass with the change.

The full `ng run fineract-backoffice-ui:unit-test` run has pre-existing failures in `web-storage.adapter`, `auth.service`, `config.service`, `institution-config.service`, `navigation-config.service`, `client-form` and `has-institution-feature.directive`. They fail the same way on `main` without this branch, and the count moves a little between runs, so they look environment-dependent rather than related. `interest-pause-form` is not among them either way. Prettier is clean on both changed files.

Closes #512

### DIFF
--- a/src/app/features/loans/interest-pauses/interest-pause-form.component.test.ts
+++ b/src/app/features/loans/interest-pauses/interest-pause-form.component.test.ts
@@ -94,6 +94,18 @@ describe('InterestPauseFormComponent', () => {
     expect(routerSpy.navigate).toHaveBeenCalledWith(['/loans', 1, 'interest-pauses']);
   });
 
+  it.each(['abc', '0', '-1', '1.5', ' '])(
+    'should not enter edit mode for the unusable variation id %p',
+    async (variationId) => {
+      await setup({ loanId: '1', variationId });
+
+      expect(component.isEditMode()).toBe(false);
+      expect(component.variationId).toBeNull();
+      expect(serviceSpy.getLoansLoanIdInterestPauses).not.toHaveBeenCalled();
+      expect(routerSpy.navigate).toHaveBeenCalledWith(['/loans', 1, 'interest-pauses']);
+    },
+  );
+
   it('should load and pre-fill the selected pause in edit mode', async () => {
     await setup({ loanId: '1', variationId: '7' }, [
       { id: 6, startDate: '2026-03-10', endDate: '2026-03-20' },
--- a/src/app/features/loans/interest-pauses/interest-pause-form.component.ts
+++ b/src/app/features/loans/interest-pauses/interest-pause-form.component.ts
@@ -48,6 +48,19 @@ import {
 } from '../../../core/utils/date-formatter';
 
 /**
+ * Parses a route segment that is meant to be a database id. Returns null for
+ * anything that cannot be one, which `+value` would otherwise turn into NaN or
+ * into a number the API cannot address.
+ */
+function toRouteId(value: string | null): number | null {
+  if (value === null || value.trim() === '') {
+    return null;
+  }
+  const id = Number(value);
+  return Number.isInteger(id) && id > 0 ? id : null;
+}
+
+/**
  * Create or edit an interest pause period on a loan. Submits a start/end date pair formatted
  * with the Fineract date format and locale. The loan and variation ids come from the route.
  */
@@ -171,14 +184,24 @@ export class InterestPauseFormComponent implements OnInit {
   ngOnInit(): void {
     const loanId = this.route.snapshot.paramMap.get('loanId');
     const variationId = this.route.snapshot.paramMap.get('variationId');
-    if (loanId) {
-      this.loanId = +loanId;
+    this.loanId = toRouteId(loanId);
+    if (variationId === null) {
+      return;
     }
-    if (variationId) {
-      this.variationId = +variationId;
-      this.isEditMode.set(true);
-      this.loadPause();
+
+    // The route puts no constraint on the segment, so `edit/abc` used to give
+    // `+'abc'` -> NaN. NaN is falsy, so `loadPause` and the save branch both
+    // skipped it while `isEditMode` stayed true: the screen said Edit, the
+    // pickers were empty, and Save created a second pause. Send an id that
+    // cannot address a pause back to the list instead.
+    this.variationId = toRouteId(variationId);
+    if (this.variationId === null) {
+      this.onCancel();
+      return;
     }
+
+    this.isEditMode.set(true);
+    this.loadPause();
   }
 
   private loadPause(): void {


Aman-Mittal (commented):
Used AI to review this

Reviewed locally against `55abe539`. The diagnosis is right, the fix is the right shape, and the tests are real — I reverted just the component and re-ran your spec:

```
5 failed | 5 passed (10)     AssertionError: expected true to be false
```

and all 10 pass with it. Deriving `isEditMode` after the id has been validated is exactly the right move: the original bug was that `isEditMode` and the two `if (!this.variationId)` guards were computed from the same segment by two different truthiness rules, and that can no longer happen.

To your question — redirecting is the right call, and it matches what #512 suggested. Staying put on an `/edit/` URL that is really a create form is the worse of the two. If you want to go one better, a brief notification on the way out would tell the user why they moved, but I would not hold the PR for it.

Two things I would like changed, and one nit.

### 1. The redirect can land on a route that does not exist

`onCancel()` navigates to `['/loans', this.loanId, 'interest-pauses']`. Now that `loanId` also goes through `toRouteId`, `/loans/abc/interest-pauses/edit/7` sets `loanId` to `null` and the redirect becomes `/loans/null/interest-pauses`. Confirmed by asserting on the spy:

```
expected [ [ '/loans', null, 'interest-pauses' ] ]
```

This is not a regression — before your change that same path produced `/loans/NaN/interest-pauses` — but your PR is the natural place to finish it, since `toRouteId` is now telling you the id is unusable. Falling back to `/loans` when `loanId` is null would close it.

### 2. An unusable loan id still leaves a dead Edit screen

With `loanId` unusable but `variationId` fine (`/loans/abc/interest-pauses/edit/7`), the component sets `isEditMode` to true, `loadPause()` returns early on `!this.loanId`, and `onSubmit` returns early on `if (!this.loanId) return;`. The result is a screen that says Edit Interest Pause, shows nothing, and does nothing at all when Save is pressed — the same shape as the bug you are fixing, one field over.

`toRouteId(loanId) === null` is already the signal; sending that case to `/loans` alongside case 1 handles both.

### 3. Nit: `%p` does not substitute here

All five parameterized cases render their title literally, so when one fails you cannot tell which id it was:

```
× should not enter edit mode for the unusable variation id %p
× should not enter edit mode for the unusable variation id %p
...
```

`%s` does substitute — verified by forcing a failure:

```
× should not enter edit mode for the unusable variation id abc
× should not enter edit mode for the unusable variation id 0
× should not enter edit mode for the unusable variation id -1
× should not enter edit mode for the unusable variation id 1.5
```

Worth having, since the five cases are the point of the table.

### About those "pre-existing failures"

They are not pre-existing — your `node_modules` has drifted from `package-lock.json`. I lost a while to exactly this today and diagnosed it the wrong way round at first. Check with:

```
node -e "console.log(require('./node_modules/@angular/build/package.json').version, require('./node_modules/vitest/package.json').version)"
```

If that prints anything other than `22.1.0 4.1.11`, `npm ci` will fix it. On the older builder, files that pass individually fail when another file is in the same run, which is what makes it look environmental.

Your branch on correct dependencies:

```
Test Files  236 passed (236)
     Tests  1424 passed (1424)
```

So the suite is clean and `interest-pause-form` is not a special case in it.

Happy to approve once 1 and 2 are in. Thanks for picking this up — the parameterized table covering `0`, `-1` and `1.5` alongside `abc` is more thorough than the issue asked for.

Aman-Mittal:
> Used AI to review this
> 
> Reviewed locally against `55abe539`. The diagnosis is right, the fix is the right shape, and the tests are real — I reverted just the component and re-ran your spec:
> 
> ```
> 5 failed | 5 passed (10)     AssertionError: expected true to be false
> ```
> 
> and all 10 pass with it. Deriving `isEditMode` after the id has been validated is exactly the right move: the original bug was that `isEditMode` and the two `if (!this.variationId)` guards were computed from the same segment by two different truthiness rules, and that can no longer happen.
> 
> To your question — redirecting is the right call, and it matches what #512 suggested. Staying put on an `/edit/` URL that is really a create form is the worse of the two. If you want to go one better, a brief notification on the way out would tell the user why they moved, but I would not hold the PR for it.
> 
> Two things I would like changed, and one nit.
> ### 1. The redirect can land on a route that does not exist
> 
> `onCancel()` navigates to `['/loans', this.loanId, 'interest-pauses']`. Now that `loanId` also goes through `toRouteId`, `/loans/abc/interest-pauses/edit/7` sets `loanId` to `null` and the redirect becomes `/loans/null/interest-pauses`. Confirmed by asserting on the spy:
> 
> ```
> expected [ [ '/loans', null, 'interest-pauses' ] ]
> ```
> 
> This is not a regression — before your change that same path produced `/loans/NaN/interest-pauses` — but your PR is the natural place to finish it, since `toRouteId` is now telling you the id is unusable. Falling back to `/loans` when `loanId` is null would close it.
> ### 2. An unusable loan id still leaves a dead Edit screen
> 
> With `loanId` unusable but `variationId` fine (`/loans/abc/interest-pauses/edit/7`), the component sets `isEditMode` to true, `loadPause()` returns early on `!this.loanId`, and `onSubmit` returns early on `if (!this.loanId) return;`. The result is a screen that says Edit Interest Pause, shows nothing, and does nothing at all when Save is pressed — the same shape as the bug you are fixing, one field over.
> 
> `toRouteId(loanId) === null` is already the signal; sending that case to `/loans` alongside case 1 handles both.
> ### 3. Nit: `%p` does not substitute here
> 
> All five parameterized cases render their title literally, so when one fails you cannot tell which id it was:
> 
> ```
> × should not enter edit mode for the unusable variation id %p
> × should not enter edit mode for the unusable variation id %p
> ...
> ```
> 
> `%s` does substitute — verified by forcing a failure:
> 
> ```
> × should not enter edit mode for the unusable variation id abc
> × should not enter edit mode for the unusable variation id 0
> × should not enter edit mode for the unusable variation id -1
> × should not enter edit mode for the unusable variation id 1.5
> ```
> 
> Worth having, since the five cases are the point of the table.
> ### About those "pre-existing failures"
> 
> They are not pre-existing — your `node_modules` has drifted from `package-lock.json`. I lost a while to exactly this today and diagnosed it the wrong way round at first. Check with:
> 
> ```
> node -e "console.log(require('./node_modules/@angular/build/package.json').version, require('./node_modules/vitest/package.json').version)"
> ```
> 
> If that prints anything other than `22.1.0 4.1.11`, `npm ci` will fix it. On the older builder, files that pass individually fail when another file is in the same run, which is what makes it look environmental.
> 
> Your branch on correct dependencies:
> 
> ```
> Test Files  236 passed (236)
>      Tests  1424 passed (1424)
> ```
> 
> So the suite is clean and `interest-pause-form` is not a special case in it.
> 
> Happy to approve once 1 and 2 are in. Thanks for picking this up — the parameterized table covering `0`, `-1` and `1.5` alongside `abc` is more thorough than the issue asked for.

Merging This we can handle the changes in separate PR

Aman-Mittal (approved):

PR #529, opened 2026-09-06
fix: redirect blank section routes

### LINKED ISSUE #527: Accounting, Organization and System are reachable from the menu and render nothing at all
## What happens

Three top-level menu destinations render a completely empty `main`:

```
/accounting     main text chars=0
/organization   main text chars=0
/system         main text chars=0
```

Measured as `document.querySelector('main').innerText.trim().length` after the route settles, against the local docker stack. No card, no list, no links, no table, no error — nothing.

The Accounting screen, empty

The URL is right, the sidebar highlights the section, the tab title is right. The page is just blank.

## Why it happens

Each is a parent route that declares a `title` and lazy-loads its children, with no component of its own and no redirect to a default child. `app.routes.ts` for accounting:

```ts
{
  path: 'accounting',
  title: 'nav.accounting',
  loadChildren: () => import('./features/accounting/accounting.routes').then((m) => m.ACCOUNTING_ROUTES),
}
```

`ACCOUNTING_ROUTES` has no `path: ''` entry, so nothing matches the bare section path.

## Why it matters

These are not obscure deep links — they are three of the twenty-five top-level groups in the navigation, and two of them (`organization`, `system`) are where an administrator starts. A new user clicks Accounting, gets a blank screen, and has no way to tell whether the application is broken, whether their role lacks a permission, or whether the page is still loading. All three look identical when nothing renders.

It also makes the section unreachable by URL: bookmarking or sharing `/accounting` gets the recipient a blank page rather than the section.

## Two ways to fix it

Either would do, and the choice is a product one:

1. **Redirect to a first child.** `{ path: '', redirectTo: 'chart-of-accounts', pathMatch: 'full' }` inside each section's routes. Cheapest, and matches what the sidebar already does when a group is expanded.
2. **Give the section a landing screen** listing what is in it, which is also somewhere to put the section's own explanation and permission-gated tiles.

Option 1 is the smaller change; option 2 is the better answer if these sections are meant to be navigable destinations rather than menu folders. Whichever is chosen, all three should be consistent with each other.

## How to check it

```
npm start
# sign in, then visit each directly
/accounting        → blank
/organization      → blank
/system            → blank
```

Or without a browser: none of the three sections' `*.routes.ts` files contains a `path: ''` entry.

```
grep -c "path: ''" src/app/features/accounting/accounting.routes.ts
```

## Environment

`main` at `55abe539`, `apache/fineract:latest` via `deploy/docker-compose-e2e.yml`, Chromium at 1440x900 and at 412x915 — blank at both widths.

Found while working on the guided tour: the tour has nothing to point at on these three screens, which is what drew attention to them. Not caused by that work, and present on `main` before it.

JebastineE:
## What and why

Add default redirects for the Accounting, Organization, and System section routes so navigating to `/accounting`, `/organization`, or `/system` opens the first available screen instead of rendering a blank page.

Closes #527

## Verification

- Added regression tests for the three section-root redirects using Angular RouterTestingHarness.
- Feature route tests: 34/34 passed.
- ESLint: 0 errors, 0 warnings.
- Prettier format check passed.
- Route permission static check passed.
- Navigation ID check passed.
- UI was not manually exercised; redirect behavior is covered by the route tests.

## Screenshots

Not applicable — this change fixes route behavior and does not introduce a new UI.

## Checklist

- [ ] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant. Not applicable; this is a route-level redirect covered by Angular router tests.

### DIFF
--- a/e2e/feature-coverage.spec.ts
+++ b/e2e/feature-coverage.spec.ts
@@ -184,6 +184,21 @@ test.describe('Navigation & Sidebar', () => {
     }
   });
 
+  test('navigating to /accounting redirects to /accounting/chart-of-accounts', async ({ page }) => {
+    await page.goto('/accounting');
+    await expect(page).toHaveURL(/\/accounting\/chart-of-accounts$/);
+  });
+
+  test('navigating to /organization redirects to /organization/offices', async ({ page }) => {
+    await page.goto('/organization');
+    await expect(page).toHaveURL(/\/organization\/offices$/);
+  });
+
+  test('navigating to /system redirects to /system/data-tables', async ({ page }) => {
+    await page.goto('/system');
+    await expect(page).toHaveURL(/\/system\/data-tables$/);
+  });
+
   test('header shows logged-in user info', async ({ page }) => {
     await expect(page.getByText(TEST_USER)).toBeVisible();
     await expect(page.getByText('Business Date:')).toBeVisible();
--- a/src/app/features/accounting/accounting.routes.test.ts
+++ b/src/app/features/accounting/accounting.routes.test.ts
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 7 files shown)


Aman-Mittal:
@JebastineE Add some e2e playwrite tests for this

Aman-Mittal (approved):

PR #532, opened 2026-09-07
fix: keep client stepper visible on mobile

### LINKED ISSUE #489: Create Client's third step is rendered off-screen and clipped away on phone viewports
## What happens

At a phone viewport, the Create Client stepper lays its three steps out in a single non-wrapping row that is wider than the space it is given. Step 3 renders past the right edge and is then clipped by an ancestor — so it is neither visible nor reachable by scrolling. A user filling in the form has no way to know a third step exists.

Measured at **368×691** on `/clients/create`, step 1 active:

```
<ol class="stepper">   scrollWidth: 442   clientWidth: 264
                       display: flex   flex-wrap: nowrap   overflow-x: visible

step labels (viewport width 368):
  "Client Type"       left  88   right 165    visible
  "Personal Details"  left 225   right 333    visible, but clipped by the card at ~332
  "Contact Details"   left 393   right 494    entirely outside the viewport
```

Create Client at 368x691 — "Personal Details" is cut off mid-word at the card edge and "Contact Details" is absent entirely

## Why it is clipped rather than scrollable

Walking the ancestor chain for `overflow-x`:

```
OL.stepper            visible
APP-STEPPER           visible
FORM.client-form      visible
ION-CARD-CONTENT      visible
ION-CARD              hidden     ← clips here
DIV.form-container    visible
MAIN.content-area     auto       ← would have scrolled, but the card already cut the content off
```

`MAIN.content-area` does allow horizontal scrolling, but `ion-card`'s `overflow-x: hidden` clips the overflow before it ever reaches that scroll container. The result is content that is neither shown nor scrollable to.

## Cause

`src/app/shared/components/stepper/stepper.component.ts` has no responsive handling at all:

```css
.stepper     { display: flex; ... }        /* no flex-wrap */
.step-label  { white-space: nowrap; ... }
```

```console
$ grep -c "@media" src/app/shared/components/stepper/stepper.component.ts
0
```

Not a single media query in the file. The layout is desktop-only by construction, and `DOCS/MOBILE.md` makes clear that phone support is an intended target rather than an edge case.

## Suggested fix

Give `.stepper` a mobile behaviour below a breakpoint. Any of these works:

- `flex-wrap: wrap` so the steps stack onto two rows;
- make the strip horizontally scrollable (`overflow-x: auto` on `.stepper` **and** relaxing the `ion-card` clip) with a visible scroll affordance;
- a compact variant that shows numbered circles only, expanding the text label for the active step.

The third degrades best — three small circles fit comfortably in 264px and still communicate "3 steps, you are on 1".

## Environment

Reproduced against a clean checkout of `main` (`a24a06ba`) served with `ng serve`, Chrome at 368×691 with mobile emulation and touch.

ryux1:
## What and why

At mobile widths, the client-creation stepper rendered every non-wrapping label in one row, while the surrounding Ionic card clipped the overflow. This keeps all numbered markers in view, gives layout space only to the active label at the shared 768px mobile breakpoint, and exposes each translated label plus the current step to assistive technology. Wider layouts keep the existing presentation.

Closes #489

## Verification

- `npm run check:responsive`
- `WATCHPACK_POLLING=true CHOKIDAR_USEPOLLING=true npm run test:e2e -- e2e/mobile-shell.spec.ts --project=mobile --grep 'keeps every client-creation step reachable'` — passed against the mocked backend with the Pixel 7 profile
- `npm test -- --watch=false` — 1,461 tests passed across 239 files
- `npm run lint`
- `npm run format:check`
- `npm run check:icons`
- `npm run i18n:check`
- `npm run build`

## Screenshots

The before state is captured in #489. The focused mobile browser test verifies the fixed geometry by asserting that the stepper does not overflow and all three markers remain within the viewport.

## Checklist

- [x] UI workflow changes include suitable e2e coverage, including real-backend testing where relevant.

### DIFF
--- a/src/app/shared/components/stepper/stepper.component.ts
+++ b/src/app/shared/components/stepper/stepper.component.ts
@@ -38,6 +38,8 @@ import { TranslatePipe } from '../../../core/adapters';
           class="step"
           [class.step-done]="i < currentIndex()"
           [class.step-active]="i === currentIndex()"
+          [attr.aria-label]="label | appTranslate"
+          [attr.aria-current]="i === currentIndex() ? 'step' : null"
         >
           <span class="step-marker">
             @if (i < currentIndex()) {
@@ -107,6 +109,26 @@ import { TranslatePipe } from '../../../core/adapters';
         height: 2px;
         margin: 0 12px;
         background: var(--border-color, #ccc);
+      }
+      @media (max-width: 600px) {
+        .step {
+          min-width: 0;
+        }
+        .step:last-child {
+          flex: 1;
+        }
+        .step-label {
+          display: none;
+        }
+        .step-active .step-label {
+          display: block;
+          min-width: 0;
+          white-space: normal;
+          overflow-wrap: anywhere;
+        }
+        .step-connector {
+          margin: 0 6px;
+        }
       }
     `,
   ],
--- a/src/app/shared/components/stepper/stepper.component.ts
+++ b/src/app/shared/components/stepper/stepper.component.ts
@@ -110,7 +110,7 @@ import { TranslatePipe } from '../../../core/adapters';
         margin: 0 12px;
         background: var(--border-color, #ccc);
       }
-      @media (max-width: 600px) {
+      @media (max-width: 768px) {
         .step {
           min-width: 0;
         }
--- a/e2e/mobile-shell.spec.ts
+++ b/e2e/mobile-shell.spec.ts
@@ -174,6 +174,38 @@ test.describe('the shell at a phone viewport', () => {
     expect(styles.labelFits).toBe(true);
   });
 
+  test('keeps every client-creation step reachable on a phone viewport', async ({ page }) => {
+    await page.route(/\/api\/v1\/offices/, (route) =>
+      route.fulfill({ status: 200, contentType: 'application/json', body: '[]' }),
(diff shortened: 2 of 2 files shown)


src/app/shared/components/stepper/stepper.component.ts
```diff
@@ -38,6 +38,8 @@ import { TranslatePipe } from '../../../core/adapters';
           class="step"
           [class.step-done]="i < currentIndex()"
           [class.step-active]="i === currentIndex()"
+          [attr.aria-label]="label | appTranslate"
```
Aman-Mittal, line 41: label | appTranslate ?


Also please share screen share and videos

ryux1: The label is already passed through `appTranslate` for both the visible step text and the `aria-label`.

I also recorded the focused Playwright test on the Pixel 7 viewport. It passes and shows all three markers inside the viewport with only the active translated label visible:

Mobile client stepper

[Video recording (WebM)](https://raw.githubusercontent.com/ryux1/fineract-backoffice-ui/evidence/pr-532-mobile-stepper/.pr-evidence/532/mobile-stepper.webm)

Aman-Mittal (approved):

PR #535, opened 2026-09-09
Send leftover audit trails path to the audits screen

### LINKED ISSUE #528: The Audit Trails menu entry silently drops the user on the dashboard
## What happens

Navigating to `/security/audit-trails` ends up on `/dashboard`. No message, no error, no indication that anything was refused or that the address was wrong:

```
requested  /security/audit-trails
landed on  /dashboard
```

Confirmed by reading `location.pathname` after the route settles, not by eyeballing it.

Where /security/audit-trails actually lands

The screenshot is the dashboard, because that is what the user gets.

## Why it matters

Audit trails are how an institution answers "who changed this, and when" — the screen a checker, an auditor or a support engineer reaches for when something is disputed. Being quietly relocated to the dashboard is worse than a plain error, because there is nothing to act on: the user cannot tell whether

- the feature does not exist in this build,
- their role is missing a permission,
- the URL they were given is stale, or
- the page failed to load.

All four look identical, and the natural response — try again — produces the same result.

The signed-in user in the reproduction holds `ALL_FUNCTIONS`, so a permission gate is not the obvious explanation, though I have not ruled out a guard that resolves to a redirect regardless.

## What to look at

I did not chase the cause, so this needs a look rather than a patch:

- whether `/security/audit-trails` is the route the sidebar entry actually points at, or whether the entry and the route have drifted apart;
- whether a route guard is redirecting rather than sending the user to the Access Denied screen the application already has for a refused route;
- whether the entry should be there at all in this build.

Whichever it is, the outcome should be one of: the screen, an Access Denied page, or no menu entry. A silent redirect to somewhere else is the one option that tells the user nothing.

## How to check it

```
npm start
# sign in as a user with ALL_FUNCTIONS, then either
#  - click Audit Trails in the sidebar, or
#  - visit /security/audit-trails directly
```

Both land on `/dashboard`.

## Environment

`main` at `55abe539`, `apache/fineract:latest` via `deploy/docker-compose-e2e.yml`, Chromium at 1440x900. Reproduced on `main` before the guided-tour branch as well, so it is not related to that work — it turned up because a tour could not be verified on a screen that never loads.

Hashim1999164:
## What and why

The audits screen lives at /security/audits. The leftover audit trails path did not match anything, so the catch all dumped people on the dashboard with no explanation.

This adds a redirect to the audits screen so that leftover URL still hits the existing READ_AUDIT check instead of looking like the feature vanished.

Fixes #528

## Verification

Ran the new security routes unit test. The leftover path now lands on /security/audits, not the dashboard.

## Screenshots

Not applicable. Routing only.

## Checklist

I did not hand edit generated files under src/app/api/.
No new adapters.
No new user facing strings.
Added a unit test for the leftover path.
No UI workflow change so no e2e.
Commits are signed.

### DIFF
--- a/src/app/features/security/security.routes.ts
+++ b/src/app/features/security/security.routes.ts
@@ -72,4 +72,11 @@ export const SECURITY_ROUTES: Routes = [
     loadComponent: () =>
       import('./audit-logs/audit-logs-list.component').then((m) => m.AuditLogsListComponent),
   },
+  {
+    // Leftover bookmarks and docs used this path. Without an alias the catch-all
+    // dumped people on the dashboard with no explanation.
+    path: 'audit-trails',
+    redirectTo: 'audits',
+    pathMatch: 'full',
+  },
 ];
--- a/e2e/feature-coverage.spec.ts
+++ b/e2e/feature-coverage.spec.ts
@@ -510,6 +510,12 @@ test.describe('Security', () => {
     await expect(page).toHaveURL('/security/audits');
     await expect(page.locator(CARD_TITLE).first()).toContainText(/Audit/i);
   });
+
+  test('leftover audit trails path lands on audits, not the dashboard', async ({ page }) => {
+    await page.goto('/security/audit-trails');
+    await expect(page).toHaveURL('/security/audits');
+    await expect(page.locator(CARD_TITLE).first()).toContainText(/Audit/i);
+  });
 });
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


src/app/features/security/security.routes.ts
```diff
@@ -72,4 +72,11 @@ export const SECURITY_ROUTES: Routes = [
     loadComponent: () =>
       import('./audit-logs/audit-logs-list.component').then((m) => m.AuditLogsListComponent),
   },
+  {
+    // Leftover bookmarks and docs used this path. Without an alias the catch-all
+    // dumped people on the dashboard with no explanation.
```
Aman-Mittal, line 77: Add playwrite tests for it

Hashim1999164: yeah added a playwright test for that leftover path

Aman-Mittal (approved):

Issue #530, opened 2026-09-06
Make the component library plug-and-play behind an adapter boundary


Aman-Mittal:
A proposal for discussion. Nothing here has been implemented.

**Goal:** swapping the component library should mean reimplementing an adapter, not editing 296 files. The same change makes the app more customizable, because whatever the app owns is what a deployment can change.

## Where we are

- `5,424` `<ion-*>` occurrences across 296 files; 299 files import `@ionic/angular/standalone`.
- About **62% of that is cosmetic** — cards, buttons, icons, grid.
- **~770 `ngModel` bindings** across six control types, with **zero custom `ControlValueAccessor`s**. Ionic currently defines what a value *is*.
- Good news: there is **no Ionic app shell** (no `ion-app`, `ion-content`, `ion-router-outlet`, `ion-menu`, `ion-tabs`). The layout is ours already.

## The catch

Wrapping components fixes the source, but the source is the cheap half.

| | Would need changing |
|---|---|
| Unit tests | ~15–20% of 237 files, mostly one-line import swaps |
| Source | 296 files, but ~62% cosmetic |
| **E2E** | **39 of 52 specs**, 16 of which bypass our helper layer entirely |

E2E asserts against rendered DOM, so wrapping components does nothing for it by itself. **The boundary has to cover the test seam too, or it is not plug and play.**

## Proposed design

`src/app/ui/` holds app-owned primitives (`app-button`, `app-select`, `app-input`, …). The public API is ours; the vendor lives inside. A swap reimplements that directory.

Three tiers, because they have different answers:

1. **Cosmetic** — card, button, icon, layout. A template swap. High volume, low risk.
2. **Form controls** — each owns its `ControlValueAccessor`. This is the important one: it moves the value contract (ISO date strings, select values) into the app, so a swap fails loudly instead of corrupting data quietly. Keeping today's selectors means ~770 bindings move without touching 296 templates.
3. **Behavioural** (select popup, tabs, overlays) — build on **`@angular/cdk`**, which `AGENTS.md` already retains "for unstyled primitives" and which `@angular/cdk/table` already uses in 28 files. CDK is headless, so this tier becomes permanently vendor-independent rather than merely swappable.

**Test seam:** each primitive exposes a stable `data-testid` and known ARIA, so `e2e/utils/` targets our contract instead of Ionic's internals. That is what makes the 39-spec number fall.

[...]

PR #533, opened 2026-09-08
feat(clients): unify client search with pagination

FenjuFu:
## What and why

Consolidate the standalone client text search into the Clients table, as requested in review. One search field supports paginated name, account-number, external-ID and mobile-number searches; `/clients/search` redirects to `/clients` and its duplicate navigation entry is removed.

Keep status filtering in the same table. The backend's status-filtered endpoint supports name matching, so that mode displays a matching placeholder and hint. Preserve supported sorting, reset unsupported Office sorting for broad text queries, and reset pagination when the query changes. Exact v2 totals such as 1, 11 and 21 render as numbers while v1 retains its unknown-total behavior.

Closes #493

AI assistance was used for implementation, regression tests, review and validation. This extends the original label correction to address the maintainer's request for one client-search section with pagination.

## Verification

- **Full fork [CI](https://github.com/FenjuFu/fineract-backoffice-ui/actions/runs/34441218209) passed on `6b8656359d9df9a9c0fa5dbcbed9a0cba1604c83`.** This includes the full application unit suites in UTC and America/New_York, microfrontend tests, ESLint, formatting, production/container builds, API-client drift, translation/navigation checks, security, GA readiness and license/RAT checks.
- **Full fork [E2E](https://github.com/FenjuFu/fineract-backoffice-ui/actions/runs/34441215137) passed on the same head:** all four mocked-backend shards, all three real-Fineract shards, mobile viewport, two-factor authentication and merged reports.
- Local evidence: 53 focused client/data-table unit regressions, 8 route-title harness tests and 26 desktop/mobile Playwright cases passed. The route harness preserves redirects and verifies the legacy URL resolves to the client-list title. Local browser checks used mocked endpoints; real-backend acceptance is covered by the fork E2E run above.
- All five PR commits have matching FenjuFu noreply sign-offs and verified GitHub signatures. The upstream [signed-commit check](https://github.com/apache/fineract-backoffice-ui/actions/runs/34441216859) passed. Changes are additive; existing history is preserved.
- **Upstream approval remains separate:** [CI](https://github.com/apache/fineract-backoffice-ui/actions/runs/34441218550), [E2E](https://github.com/apache/fineract-backoffice-ui/actions/runs/34441218627) and [CodeQL](https://github.com/apache/fineract-backoffice-ui/actions/runs/34441218504) still require maintainer approval. Fork results do not replace those upstream required checks.

## Screenshots

Captured from the final code with mocked client data.

<details>
[...]

PR #540, opened 2026-09-10
feat(ui): establish component boundary with accessible tabs

FenjuFu:
## What and why

Shared custom-field tabs now use an app-owned component and test contract. Arrow keys move focus without fetching a table; Enter/Space activates the selected table. Native buttons and CDK handle keyboard navigation, disabled items and RTL, while the feature receives a table-name string and owns loading the panel.

This also establishes the first migration increment from #530: ADR 0005 specifies the complete component/CVA value, keyboard/ARIA, browser-helper, theming and rollout contracts; the Ionic import ratchet records existing use while rejecting new unsuppressed dependencies outside UI implementations. Existing Material/i18n restrictions and other-rule allowances are preserved. AGENTS.md and STYLE.md direct new feature/shared code through app/ui.

Refs #530. Remaining tab strips, popup controls, form CVAs, cosmetic primitives, deployment-token expansion and legacy browser helpers remain under the overall migration issue. This PR delivers the boundary and one independently verifiable behavioural primitive; it does not close that issue.

## Verification

- Final head: `e992e9b88315455bcd310dc6941946e3804f7e49`.

- Local: 17 rendered Angular unit/integration tests and 3 ESLint boundary tests passed. App/E2E TypeScript checks, changed-source ESLint, formatting and git diff checks passed. Other lint-rule allowances were compared with the base and preserved.

- Full fork [CI](https://github.com/FenjuFu/fineract-backoffice-ui/actions/runs/34444465337) passed on this head, including full unit suites, full lint/suppression pruning, production/container builds, API drift, translations, security, GA and license/RAT checks.

- Full fork [E2E](https://github.com/FenjuFu/fineract-backoffice-ui/actions/runs/34444461409) passed on the same head: all mocked and real-Fineract shards, mobile and two-factor authentication. The new custom-field cases exercise the app-owned keyboard/panel contract at 1280px and 390px; they use mocked group data and attach screenshots after loading completes.

- All three PR commits have matching FenjuFu noreply sign-offs and valid GitHub signatures. No generated API files were hand-edited. The upstream [signed-commit check](https://github.com/apache/fineract-backoffice-ui/actions/runs/34445927919) passed.
- Upstream [CI](https://github.com/apache/fineract-backoffice-ui/actions/runs/34445928021), [E2E](https://github.com/apache/fineract-backoffice-ui/actions/runs/34445928047) and [CodeQL](https://github.com/apache/fineract-backoffice-ui/actions/runs/34445928007) still require maintainer approval. The fork results above are independent evidence.

## Screenshots

Unmodified image attachments from the passing new CI browser cases, showing the custom-field region after the panel finishes loading. Screenshots are stored on a separate review branch.

<details>

<summary>Desktop, 1280px viewport</summary>

App-owned custom-field tabs on desktop

</details>

<details>

<summary>Narrow, 390px viewport</summary>

App-owned custom-field tabs on a narrow viewport

</details>

## AI assistance (optional)

- Tool: Codex.

- Harness / workflow: assisted implementation and review, targeted local checks, and full fork GitHub Actions validation.

## Checklist

- [x] I did not hand-edit generated files under src/app/api/.

- [x] Feature translation uses the core adapter; DOM/focus behaviour is confined to the app/ui primitive.

- [x] User-facing static strings use translation keys; registered table names remain user data.

- [x] Added rendered unit, import-boundary and browser contract tests.

- [x] UI changes have browser coverage; full mocked and real-backend suites passed.

- [x] Commits are signed and contain matching sign-offs.

- [x] Followed the [AI-assisted contributions guidance](https://github.com/apache/fineract-backoffice-ui/blob/main/CONTRIBUTING.md#ai-assisted-contributions).

### DIFF
--- a/DOCS/adr/0005-ui-boundary.md
+++ b/DOCS/adr/0005-ui-boundary.md
@@ -1,0 +1,119 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# ADR 0005: An application-owned UI and test boundary
+
+- **Status:** Proposed; first tab migration implemented for review
+- **Date:** 2026-09-10
+- **Discussion:** [#530](https://github.com/apache/fineract-backoffice-ui/issues/530)
+- **Scope:** One UI implementation at a time; incremental migration, not a second component library
+
+## Problem and decision
+
+ADR 0003 isolates imperative dependencies but deliberately leaves template components outside
+the boundary. A library swap still changes hundreds of templates, form-value semantics, and
+browser locators. This proposal extends that boundary to `src/app/ui/` and to the test contract.
+It does not declare the overall migration complete.
+
+Features import app-owned components from `src/app/ui/`; their public inputs, outputs, projected
+content, ARIA and value contracts name application concepts. Ionic may be used inside this
+directory while a primitive is being implemented. Behavioural primitives use CDK and native
+elements where possible. Existing OVERLAY, I18N, STORAGE and DOWNLOAD adapters remain in force.
+No dependency is added and no second vendor runs alongside Ionic.
+
+## Public contracts for the whole boundary
+
+| Tier                                       | App-owned contract                                                                                                                                                                                                 | Acceptance before migrating consumers                                                                                                                                                                                            |
+| ------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Behavioural: tabs, popups, dialogs         | Values and dismissal reasons, focus entry/return, keyboard behaviour, roles and accessible names; no vendor events or controller handles                                                                           | Keyboard-only and pointer operation; disabled/empty/dynamic states; nested overlays; focus restored after close; browser checks at desktop and narrow widths                                                                     |
+| Form: input, select, date, checkbox/toggle | Angular ControlValueAccessor; `writeValue` never emits; disabled/touched/validation state; select preserves primitive identifiers; dates are ISO calendar strings, not UTC instants; explicit null/empty semantics | Shared CVA contract tests with reactive forms and ngModel; programmatic writes, reset, blur, disabled state; number/string identity and negative-timezone date round trips; the same browser helper against both implementations |
+| Cosmetic: cards, buttons, icons, layout    | Intent, label, disabled/busy state, content slots and app design tokens; buttons declare submit versus ordinary action                                                                                             | Accessible names, form submission, projected content, theme/density and responsive visual checks                                                                                                                                 |
+
+The form migration uses explicit value types: text is `string` (empty is `''`), numeric
+input is `number | null`, a single select is `string | number | null` without coercion,
+a multi-select is a readonly array of those non-null identifiers, and checkbox/toggle values
+are booleans. Calendar dates are valid `YYYY-MM-DD | null`; adapters must not round-trip them
+through UTC. Defaults, reset, required validation and backend conversion remain feature
+responsibilities. Every CVA shares disabled, blur/touched, `aria-invalid`, error-description
+and label tests; replacing the renderer must not change submitted payloads.
+
+New primitives use `app-*` selectors. Existing `ion-*` selectors remain only in unmigrated
+templates; no compatibility directive masquerades as an Ionic component. A consumer migration
+updates its import, template and test helper together. The ~770 form bindings are migrated by
+control type with contract tests, rather than being silently reinterpreted by a selector alias.
+
+Tests consume roles, accessible names, selected/expanded/disabled state and stable `data-testid`
+scopes. They never inspect vendor shadow DOM, CSS classes or `CustomEvent.detail`. Helpers in
+`e2e/utils/ui-locators.ts` are the new seam. Existing Ionic helpers stay for unmigrated controls;
+do not extend them for app-owned primitives. Stable hooks also apply to guided tours.
+
+Theme values flow from application design tokens into UI implementations. Primitives do not
+export Ionic CSS variables as public API. Deployment overrides retain the existing validated
+branding allowlist until each additional token has a documented type, fallback, accessibility
+constraint and validation test; this change does not remove that safety boundary. Dark mode,
+focus visibility, touch target size and overflow are part of each primitive's acceptance.
+
+## First independently verifiable increment
+
+`TabsComponent` is implemented without Ionic, using native buttons and CDK FocusKeyManager.
+Its `UiTab` values are strings; labels are translated by the caller. `value` is controlled by
+the feature and `valueChange` carries only a new enabled identifier. Arrow keys wrap and skip
+disabled tabs; Home/End work; RTL reverses horizontal navigation. Focus movement is separate
+from activation: Enter/Space or a click requests a value change. Tab leaves the strip normally.
+This avoids issuing data requests for every arrow key on a network-backed panel.
+
+Callers provide a stable, page-unique `idPrefix` and a tablist label. Each button has role `tab`,
+`aria-selected`, a roving tabindex and `aria-controls`. The caller renders one shared panel with
+role `tabpanel`, `tabindex=0`, `panelId()` and `aria-labelledby=tabId(selectedValue)`. This makes
+loading and panel lifecycle explicit rather than eagerly mounting every feature. A missing
+selected value leaves an enabled tab focusable, but never emits a fabricated selection.
+
+`EntityDatatablesComponent` is the first consumer. It sends the selected table name directly to
+its existing data-loading logic, with no Ionic event cast, and preserves the
+`entity-datatables-tabs` scope. Its rendered integration test and a group-detail browser test
+exercise both data loading and the ARIA panel relationship. The browser test also proves arrow
+navigation does not fetch a table until activation, and uses the app helper to switch back.
+
+## Enforcement and remaining rollout
+
+`no-restricted-imports` now rejects new `@ionic/angular` imports outside `src/app/ui/**`.
+Existing direct imports are seeded once in `eslint-suppressions.json`. CI's existing
+`--prune-suppressions` removes obsolete allowances. The UI exception permits Ionic only;
+Material and direct ngx-translate imports remain forbidden. Existing imperative adapters and
+composition roots (`app.config.ts`, test setup) retain their narrow architectural roles.
+The ESLint contract tests prove these boundaries with unsuppressed new-file examples.
+
+After review of this increment:
+
+1. Migrate the remaining entity/tab strips to the tab contract, updating guided tours and
+   Ionic-specific test hooks in the same change. Add per-screen acceptance tests where tabs
+   have conditional visibility or load data asynchronously.
+2. Move popup selection and remaining behavioural controls behind app contracts/CDK. Keep
+   the existing OVERLAY interface; preserve close reasons, focus return and backdrop behaviour.
+3. Implement each form CVA with the shared tests above, then migrate source and browser helpers
+   together. Demonstrate a second implementation in the contract fixture before calling a form
+   primitive replaceable; do not ship two libraries in production.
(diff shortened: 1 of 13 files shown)


eslint-suppressions.json
```diff
@@ -41,67 +41,102 @@
   },
   "src/app/features/accounting/accounting-closure-form.component.ts": {
     "no-restricted-imports": {
-      "count": 1
+      "count": 2
```
Aman-Mittal: Why increase count

src/app/ui/tabs/tabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+        border-bottom: 1px solid var(--border-color);
+      }
+      button {
+        flex: 0 0 auto;
+        min-height: 44px;
+        padding: var(--space-3) var(--space-4);
+        border: 0;
+        border-bottom: 2px solid transparent;
+        background: transparent;
+        color: var(--text-secondary);
```
Aman-Mittal, line 82: `--text-secondary` is not defined anywhere in the repository — this is its only occurrence. The declaration is therefore invalid at computed-value time, so `color` falls back to `inherit` and unselected tabs render at full `--text-color` rather than the muted tone intended here. The de-emphasis never appears in either theme.

The repo's token for this is `--text-muted` (`src/styles/_common.scss:33`, redefined for dark at `:92`).

```suggestion
        color: var(--text-muted);
```

eslint.config.js
```diff
@@ -252,6 +242,20 @@ module.exports = tseslint.config(
...
+    // UI implementations may name their vendor, but retain the Material and i18n boundaries.
+    files: ['src/app/ui/**/*.ts', 'src/app/testing/ionic-testing.ts'],
+    rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: restrictedImportPatterns.filter(
+            (pattern) => !pattern.group.includes('@ionic/angular'),
+          ),
+        },
```
Aman-Mittal, line 255: This filter drops the entire `@ionic/angular` pattern for `src/app/ui/**` and `src/app/testing/ionic-testing.ts`, which also removes ADR 0003's ban on Ionic's **imperative controllers** in those paths. Before this PR the controller restriction applied to every file with no exemption.

Confirmed by running ESLint against the new config:

```
src/app/ui/probe.ts       <- ModalController   from @ionic/angular/standalone: ALLOWED
src/app/ui/probe.ts       <- AlertController   from @ionic/angular/standalone: ALLOWED
src/app/ui/probe.ts       <- LoadingController from @ionic/angular:            ALLOWED
src/app/features/probe.ts <- ModalController   from @ionic/angular/standalone: BLOCKED
```

ADR 0005 states "Imperative controllers still use OVERLAY" and "Existing OVERLAY, I18N, STORAGE and DOWNLOAD adapters remain in force", so the config is now weaker than the decision it implements. A UI primitive that needs a modal should still go through the OVERLAY adapter rather than reaching for `ModalController`.

Suggest narrowing the exemption to components only, rather than removing the Ionic entry — keep the controller-scoped `importNames` restriction that the old config had:

```js
patterns: [
  ...restrictedImportPatterns.filter((pattern) => !pattern.group.includes('@ionic/angular')),
  {
    group: ['@ionic/angular', '@ionic/angular/*'],
    importNames: [
      'ModalController',
      'ToastController',
      'AlertController',
      'LoadingController',
      'ActionSheetController',
      'PopoverController',
    ],
    message:
      "Use the OVERLAY adapter from 'app/core/adapters' instead of Ionic's controllers, inside src/app/ui as well. See DOCS/adr/0003-adapter-boundary.md.",
  },
],
```

scripts/ui-boundary.test.mjs
```diff
@@ -0,0 +1,62 @@
...
+  );
+  assert.equal(
+    (await importErrors('src/app/ui/probe.ts', '@angular/material/button', 'MatButton')).length,
+    1,
+  );
+  assert.equal(
+    (await importErrors('src/app/ui/probe.ts', '@ngx-translate/core', 'TranslateService')).length,
+    1,
+  );
+});
```
Aman-Mittal, line 53: The test name promises UI implementations "cannot bypass other adapter boundaries", but it only asserts Material and ngx-translate — which is why the controller gap in `eslint.config.js` goes unnoticed. Worth closing the loop here so the assertion matches the name:

```js
  assert.equal(
    (await importErrors('src/app/ui/probe.ts', '@ionic/angular/standalone', 'ModalController'))
      .length,
    1,
  );
```

This currently returns `0`. With the `importNames` entry restored in the `src/app/ui/**` override it returns `1`, and the OVERLAY boundary stays enforced across the whole tree as the migration adds primitives that need dialogs.

src/app/ui/tabs/tabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+        font: inherit;
+        cursor: pointer;
+      }
+      button[aria-selected='true'] {
+        color: var(--primary-color);
+        border-bottom-color: var(--primary-color);
+      }
```
Aman-Mittal, line 89: The selected tab's **label** is body-sized text in `--primary-color`. On `--card-bg` that is `#3498db` on `#ffffff` ≈ 3.2:1, short of the 4.5:1 WCAG AA minimum for body text. Dark theme is fine (`#3498db` on `#1e1e1e` ≈ 5.2:1). The `border-bottom-color` on the next line is a non-text indicator at 3:1 and is fine as-is.

The token comment in `_common.scss` does say "Keep `--primary-color` for accents, borders and text on light backgrounds", but that clause sits directly beneath its own measurement of the same ratio at 3.15:1 — it holds for accents and borders, not for text.

Worth flagging that `--primary-strong` is **not** a drop-in fix: it is not redefined in the `[data-theme='dark']` block, and `#2471a3` on `#1e1e1e` is ≈3.2:1, so a blanket swap trades the light-theme failure for a dark-theme one. This needs a themed token, the way `--guidance-highlight-color` is already themed in `_common.scss` for exactly this reason ("one colour cannot serve both").

Given ADR 0005 makes "dark mode, focus visibility, touch target size and overflow ... part of each primitive's acceptance", the first primitive seems like the right place to establish that token.

src/app/ui/tabs/tabs.component.ts
```diff
@@ -0,0 +1,148 @@
...
+  template: `
+    <div role="tablist" [attr.aria-label]="label()" data-testid="ui-tabs">
+      @for (tab of tabs(); track tab.value; let index = $index) {
+        <button
+          #tabButton
+          type="button"
+          role="tab"
+          data-testid="ui-tab"
+          [id]="tabId(tab.value)"
+          [attr.aria-controls]="panelId()"
```
Aman-Mittal, line 51: Minor, on edge states in the ARIA contract:

`aria-controls` is emitted unconditionally, but the consumer renders the `role="tabpanel"` element only under `@if (activeTable(); as dt)`. Before a selection resolves — and in `EntityDatatablesComponent` the tabs render as soon as `datatables()` is non-empty — every tab points at an ID that is not in the document.

Two related cases, both reachable through the same path: when `value()` matches no tab, no button carries `aria-selected="true"`; and when every tab is disabled, `focusIndex` is `-1` so the strip has no tab stop at all (asserted as intended at the end of the unit test). APG expects a tablist to keep exactly one tab stop and one selected tab, so a screen-reader user can land on the strip and find out where they are.

`EntityDatatablesComponent` also sets `activeTable` to `data[0]` unconditionally while `tableTabs` filters on `registeredTableName` — so a first table without a name yields a rendered panel whose `aria-labelledby` resolves to nothing, with no tab selected. Narrow, but it is the same class of gap and the non-null assertions in the template already acknowledge the field is optional.

DOCS/adr/0005-ui-boundary.md
```diff
@@ -0,0 +1,119 @@
...
+
+New primitives use `app-*` selectors. Existing `ion-*` selectors remain only in unmigrated
+templates; no compatibility directive masquerades as an Ionic component. A consumer migration
+updates its import, template and test helper together. The ~770 form bindings are migrated by
+control type with contract tests, rather than being silently reinterpreted by a selector alias.
+
+Tests consume roles, accessible names, selected/expanded/disabled state and stable `data-testid`
+scopes. They never inspect vendor shadow DOM, CSS classes or `CustomEvent.detail`. Helpers in
+`e2e/utils/ui-locators.ts` are the new seam. Existing Ionic helpers stay for unmigrated controls;
+do not extend them for app-owned primitives. Stable hooks also apply to guided tours.
```
Aman-Mittal, line 64: On "Stable hooks also apply to guided tours" — worth recording the concrete debt here, since it is invisible from this diff.

`src/app/core/services/guidance.service.ts:63` sets `TAB_GROUP_SELECTOR = 'ion-segment'`, used by five tours. Nothing breaks in this PR: the record-view outer strips are still `ion-segment` and match ahead of the nested custom-fields strip in DOM order. But the doc comment above it now reads falsely —

> All sixteen components with a tab strip render an `ion-segment`, so that is what the step points at.

That comment exists precisely because the previous selector (`.tab-group`) silently matched nothing after the Material port and four tours pointed at empty space for some time. The same failure mode returns the moment rollout step 1 migrates a record-view strip — silently, since a tour step that matches nothing does not fail CI.

Either update the comment now to say the selector is mid-migration, or promote the tours to a `data-testid` hook as part of step 1 rather than after it.

src/app/ui/tabs/tabs.component.ts
FenjuFu: Good catch, thanks. Fixed in c233ae5: unselected tabs now use `--text-muted` (5.74:1 light, 7.69:1 dark). Disabled tabs no longer share that colour; they are dimmed with opacity instead, so the two states stay distinguishable.

eslint.config.js
FenjuFu: Agreed, the override was broader than ADR 0005 allows. c233ae5 keeps the component exemption but adds back the controller-scoped `importNames` entry for `src/app/ui/**` and `ionic-testing.ts`, as you suggested. `src/app/ui` and the testing helper still lint clean with it.

scripts/ui-boundary.test.mjs
FenjuFu: Added in c233ae5. The test now asserts `ModalController` from both `@ionic/angular` and `@ionic/angular/standalone` is rejected in `src/app/ui/probe.ts`. It returned 0 before the ESLint change and returns 1 after.

src/app/ui/tabs/tabs.component.ts
FenjuFu: Thanks for measuring both themes. I added a themed `--primary-text` token in `_common.scss`: `#2471a3` in light (5.30:1 on `#ffffff`) and `#3498db` in dark (5.29:1 on `#1e1e1e`), with the measurements in the token comment. The selected label uses it; the border indicator stays on `--primary-color`.

src/app/ui/tabs/tabs.component.ts
FenjuFu: Addressed all three in c233ae5:

- `aria-controls` is emitted only on the selected tab, so no tab points at a panel that isn't rendered. When `value` matches no tab, no tab has it at all (new unit test).
- Disabled tabs use `aria-disabled` instead of `disabled`, so they stay focusable and the strip always keeps exactly one tab stop, including when every tab is disabled. Arrow keys still skip them, and click/Enter is ignored. I updated the last unit test to cover this.
- `EntityDatatablesComponent` now selects the first table with a `registeredTableName` and only renders the strip when there is at least one named table. A new test checks that the selected tab and panel point at each other.

DOCS/adr/0005-ui-boundary.md
FenjuFu: Recorded in both places in c233ae5. The `TAB_GROUP_SELECTOR` comment now says the selector is mid-migration: only the nested custom-fields strip is `app-tabs`, and migrating a record-view strip must move the selector in the same change. ADR 0005 rollout step 1 lists this as known debt, including the point that an unmatched tour step does not fail CI. Should the tours move to the `ui-tabs` test id in rollout step 1, or would you rather take that as a separate PR?

Aman-Mittal:
@FenjuFu please resolve conflicts

FenjuFu:
@Aman-Mittal Conflicts are resolved. The branch is rebased onto current `main` (1d9e686, which includes #533), and all 4 commits are signed again.

There were two conflicts:

- `src/styles/_common.scss`: `main` now defines `--primary-text` with the values this PR used (`#2471a3` light, `#3498db` dark), plus `--warning-text` and `--success-text`. I kept `main`'s definitions and dropped this PR's duplicate, so the selected tab label still resolves to the same colours.
- `eslint-suppressions.json`: merged both sides, then ran `npm run lint:prune`. The prune dropped stale entries such as `client-search-v2.component.ts`, which #533 removed. `npm run lint` passes with no unsuppressed errors.

Checks I ran locally on the new head `11197fa`: `lint:prune`, `test:scripts` (34/34), `test:eslint-rules` (26/26), `typecheck:e2e`, `i18n:check` and `test:unit`. In the full `test:unit` run, 1488 of 1489 tests passed. The one failure was `client-form.component.test.ts`, which this PR doesn't touch, and it passes 8/8 when run alone, so it looks like a timeout under load.

My earlier question about the tours is still open: should they move to the `ui-tabs` test id in rollout step 1, or in a separate PR?

Aman-Mittal (approved):

PR #544, opened 2026-09-11
fix(clients): keep date picker IDs unique

### LINKED ISSUE #541: Sometime i cannot input date time while creating client
### What happened

I cannot create Client again as i try to put date in this

Image

### Steps to reproduce

1. Create a client.
2. After Success create a another client with different legal type 

### Screen or route

_No response_

### Browser console and network errors

```text
[Ionic Error]: [ion-datetime-button] - No ion-datetime instance found for ID 'submittedOnDate-picker'. <ion-datetime-button _ngcontent-ng-c3516409172 datetime=​"submittedOnDate-picker" class=​"ion-color ion-color-primary md hydrated" color=​"primary">​flexslot#shadow-root (open)</ion-datetime-button>​
_n @ chunk-OU2VS4HL.js:29
(anonymous) @ _ionic_angular_standalone.G6ianXD1_G.js:1082
(anonymous) @ chunk-JHI3MBHO.js:25
W @ polyfills-Z57ITQZZ.js:2
v @ chunk-JHI3MBHO.js:15
componentWillLoad @ _ionic_angular_standalone.G6ianXD1_G.js:1080
G @ chunk-OU2VS4HL.js:513
on @ chunk-OU2VS4HL.js:500
n @ chunk-OU2VS4HL.js:498
(anonymous) @ chunk-OU2VS4HL.js:499
invokeTask @ polyfills-Z57ITQZZ.js:1
runTask @ polyfills-Z57ITQZZ.js:1
Oe @ polyfills-Z57ITQZZ.js:1
invokeTask @ polyfills-Z57ITQZZ.js:1
ne @ polyfills-Z57ITQZZ.js:1
N @ polyfills-Z57ITQZZ.js:1
Te @ polyfills-Z57ITQZZ.js:1
chunk-OU2VS4HL.js:29 [Ionic Error]: [ion-datetime-button] - No ion-datetime instance found for ID 'activationDate-picker'. <ion-datetime-button _ngcontent-ng-c3516409172 datetime=​"activationDate-picker" class=​"ion-color ion-color-primary md hydrated" color=​"primary">​flexslot#shadow-root (open)</ion-datetime-button>​
_n @ chunk-OU2VS4HL.js:29
(anonymous) @ _ionic_angular_standalone.G6ianXD1_G.js:1082
(anonymous) @ chunk-JHI3MBHO.js:25
W @ polyfills-Z57ITQZZ.js:2
v @ chunk-JHI3MBHO.js:15
componentWillLoad @ _ionic_angular_standalone.G6ianXD1_G.js:1080
G @ chunk-OU2VS4HL.js:513
on @ chunk-OU2VS4HL.js:500
n @ chunk-OU2VS4HL.js:498
(anonymous) @ chunk-OU2VS4HL.js:499
invokeTask @ polyfills-Z57ITQZZ.js:1
runTask @ polyfills-Z57ITQZZ.js:1
Oe @ polyfills-Z57ITQZZ.js:1
invokeTask @ polyfills-Z57ITQZZ.js:1
ne @ polyfills-Z57ITQZZ.js:1
N @ polyfills-Z57ITQZZ.js:1
Te @ polyfills-Z57ITQZZ.js:1

```

### Which backend

Local docker stack (deploy/docker-compose-e2e.yml)

### Fineract version

_No response_

### Browser and version

_No response_

### Before submitting

- [x] I searched existing issues for this.
- [x] This is a bug in the UI, not in the Fineract platform.

WilliamK112:
## What changed

- assign each client-form instance a stable unique suffix for its submitted, activation, and date-of-birth picker IDs
- bind every `ion-datetime-button` to the matching generated `ion-datetime` ID while preserving the existing test IDs
- add a regression test that creates two client-form instances and verifies all six picker IDs are distinct and the rendered button/picker pairs match

## Why

Ionic navigation can keep the previous client form alive, and these modals deliberately use `keepContentsMounted`. Reopening the create form therefore left duplicate fixed IDs in the document. `ion-datetime-button` could then resolve the wrong or missing picker, producing the reported `No ion-datetime instance found` error. This change removes that collision without changing date values or submission behavior.

## Validation

- `npm run lint`
- `npm run format:check`
- `npm run i18n:check`
- `npm run check:icons`
- `npm run check:a11y-names`
- `npm run api:surface`
- `NODE_OPTIONS=--no-experimental-webstorage npm test -- --watch=false` — 241 files, 1465 tests passed
- `npm run build`
- `bash scripts/check-license.sh`

Fixes #541

## AI assistance

Codex assisted with root-cause analysis, implementation, tests, and validation. I reviewed the final diff and verified the behavior locally.

### DIFF
--- a/src/app/features/clients/client-form.component.test.ts
+++ b/src/app/features/clients/client-form.component.test.ts
@@ -146,6 +146,34 @@ describe('ClientFormComponent', () => {
       expect(routerSpy.navigate).toHaveBeenCalledWith(['/clients']);
     });
 
+    it('uses unique date picker IDs across client form instances', () => {
+      const secondFixture = TestBed.createComponent(ClientFormComponent);
+
+      const datetimeReferences = (element: HTMLElement) =>
+        Array.from(element.querySelectorAll('ion-datetime-button')).map(
+          (button) => (button as HTMLIonDatetimeButtonElement).datetime,
+        );
+      const datetimeIds = (element: HTMLElement) =>
+        Array.from(element.querySelectorAll('ion-datetime')).map((picker) => picker.id);
+      const firstPickerIds = [
+        component.submittedOnDatePickerId,
+        component.activationDatePickerId,
+        component.dateOfBirthPickerId,
+      ];
+      const secondComponent = secondFixture.componentInstance;
+      const secondPickerIds = [
+        secondComponent.submittedOnDatePickerId,
+        secondComponent.activationDatePickerId,
+        secondComponent.dateOfBirthPickerId,
+      ];
+
+      expect(datetimeReferences(fixture.nativeElement)).toEqual(firstPickerIds);
+      expect(datetimeIds(fixture.nativeElement)).toEqual(firstPickerIds);
+      expect(new Set([...firstPickerIds, ...secondPickerIds]).size).toBe(6);
+
+      secondFixture.destroy();
+    });
+
     it('shows the wizard stepper and only the first step initially', () => {
       expect(component.showWizard()).toBe(true);
       expect(component.currentStep()).toBe(0);
--- a/src/app/features/clients/client-form.component.ts
+++ b/src/app/features/clients/client-form.component.ts
@@ -60,6 +60,8 @@ import {
   toIsoDate,
 } from '../../core/utils/date-formatter';
 
+let nextClientFormId = 0;
+
 @Component({
   selector: 'app-client-form',
   standalone: true,
@@ -187,11 +189,11 @@ import {
               <!-- Submitted On Date -->
               <ion-item fill="outline" [appTooltip]="'HELP.SUBMITTED_ON_DESC' | translate">
                 <ion-label position="stacked">{{ 'COMMON.SUBMITTED_ON' | translate }}</ion-label>
-                <ion-datetime-button datetime="submittedOnDate-picker"></ion-datetime-button>
+                <ion-datetime-button [datetime]="submittedOnDatePickerId"></ion-datetime-button>
                 <ion-modal [keepContentsMounted]="true">
                   <ng-template>
                     <ion-datetime
-                      id="submittedOnDate-picker"
+                      [id]="submittedOnDatePickerId"
                       data-testid="submittedOnDate-picker"
                       presentation="date"
                       name="submittedOnDate"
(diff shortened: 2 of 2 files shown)


Aman-Mittal:
Thanks for picking this up. The instinct to make picker IDs unique is reasonable, but I reproduced #541 in a real browser and tested this branch's approach against it — it doesn't fix the bug. Sharing the evidence, because it points somewhere different than the PR description assumes.

Repro: log in, create a client through the wizard, then start a second create with a different legal form, and inspect the `ion-datetime-button` elements on step 2.

**1. There are no duplicate IDs at the point of failure.**

On `main`, when the second form fails, the document contains exactly one `app-client-form` and one of each picker:

```
datetime ids in document: ["submittedOnDate-picker", "activationDate-picker"]
error: [ion-datetime-button] - No ion-datetime instance found for ID 'submittedOnDate-picker'
```

The previous form is fully destroyed before the new one mounts, so nothing collides. The cause is ordering, not collision: `ion-datetime-button` resolves its target exactly once in `componentWillLoad` via `document.getElementById`, and gives up permanently if that lookup misses. On a repeat visit the button initializes *before* `ion-modal`'s `keepContentsMounted` content is in the DOM. The first visit works only because the lazy chunk load happens to delay the button past that point.

**2. With this branch's approach applied, the pickers are still broken.**

Same repro, unique IDs in place — the buttons render blank and stay unusable:

```
buttons: [{"datetime":"submittedOnDate-picker-2","shadowText":"","resolves":true}, ...]
error: [ion-datetime-button] - An ID associated with an ion-datetime instance is required to function properly.
```

Moving `datetime="…"` to `[datetime]="…"` means Angular sets the property during change detection, so at `componentWillLoad` the value is still undefined and the button bails on the *first* guard instead of the second. `resolves: true` confirms the ID pairing itself is correct — the button simply never binds. This is worse for triage than the original: the error string from the issue no longer appears, so it reads as fixed while the user still cannot enter a date.

**3. The added test cannot catch this.**

It asserts that six generated ID strings are distinct and that the bound properties match. No Ionic element initializes in that environment, so it passes whether or not the picker works. The green E2E run doesn't cover it either — no existing spec opens the create form twice.

**What does work**

Leaving the existing hardcoded IDs alone and deferring the button by one render, so the `ion-datetime` is in the DOM before the button initializes — a `pickersReady` signal set from `afterNextRender`, with the `ion-datetime-button` elements behind `@if (pickersReady())`. Verified against the same repro:

```
buttons: [{"datetime":"submittedOnDate-picker","shadowText":"Sep 12, 2026","resolves":true}, ...]
```

Zero Ionic errors. That's the minimal change that addresses the actual cause.

Unique IDs are still worth doing on their own merits — 97 `ion-datetime-button` usages across 62 components reuse a small pool of names (`transactionDate-picker` in 8 files, `submittedOnDate-picker` in 7), so two such pages alive at once would collide. But that is a separate, currently-latent problem, not this one, and it would be better as its own change with a test that can actually fail.

Since this one is a blocker I'll push a fix along the lines above shortly. Happy to have you take the unique-ID work as a follow-up if you're interested — and a regression spec that opens the create form twice would be a genuinely valuable addition either way.

PR #547, opened 2026-09-12
fix(ui): keep date pickers usable on every routed form visit

Aman-Mittal:
## What changed

- add `createPickersReady()` in `src/app/shared/utils/pickers-ready.ts`
- hold every `ion-datetime-button` behind `@if (pickersReady())` across the **50 routed components** that carry the pattern — 72 buttons
- add an e2e spec that enters the office and staff forms three times each without a reload, registered to run at **both the desktop and mobile viewports**

Picker IDs are unchanged. The client form is fixed separately in #546; this PR leaves it alone to avoid a conflicting edit.

## Why

`ion-datetime-button` resolves its target `ion-datetime` exactly once, in `componentWillLoad`, through a global `getElementById`, and gives up for good when that lookup misses. The pickers live inside `ion-modal[keepContentsMounted]`, whose contents Angular mounts later in the change-detection pass.

On a first visit to a routed page the button's lazy Ionic chunk is still loading, which delays it past that point. On a revisit the chunk is cached, so the button initializes first, finds nothing, and renders a blank control that never opens. Reaching **any** of these forms a second time left every date field unusable until a full page reload — #541 is the client-form instance of it, reported because that is the busiest path.

## Scope, measured rather than assumed

- **51 routed components, 75 buttons** carry the defect. Confirmed by browser repro on two unrelated pages: the client form (#541) and `/organization/offices/create`, which renders its opening-date picker on the first visit and a dead blank control on every visit after. No submit required — Cancel and re-enter is enough. 50 components / 72 buttons are in this PR; the client form's 3 are in #546.
- **11 dialog components are not affected** and are left untouched. Opening `create-office-dialog` three times in a row renders correctly each time and logs nothing: an overlay mounts its contents before the button initializes. Verified, not inferred.

## Testing

- `npm run build`, `npm run lint`, `npm run test:unit` (241 files, 1464 tests passing)
- `npm run i18n:check`, `npm run check:icons`, `npm run check:a11y-names`, `bash scripts/check-license.sh`
- `npx playwright test date-picker-revisit --project=mocked` and `--project=mobile` — 2 passing each
- With the component changes stashed, the spec fails at **both** viewports with `visit 2 left a picker unbound`, so the coverage is real in both

`npm run format:check` reports pre-existing warnings under `_scratch/` and `LIVE_UI_BUGS.md`; no file in this change is listed.

The spec reads each button's shadow root rather than its `textContent`, which is always empty and cannot tell a bound control from a dead one.

## Notes for review

- `DUAL_VIEWPORT_SPECS` gains the glob `**/*date-picker-revisit.spec.ts` rather than a filename, so it also picks up the client-form spec in #546. Both PRs make that identical one-line edit, which merges cleanly in either order.
- Two loose ends found while sweeping, deliberately not folded in: `loanMenu-trigger` is used as an `ion-popover` trigger ID by both `loan-view` and `wc-loan-view` (latent — overlays resolve `trigger` through the same one-shot `getElementById`, so two live pages would collide), and `loans/bulk-reassignment/bulk-reassignment.component.ts` appears to be dead code, referenced only by its own test while the routed component is `bulk-loan-reassignment.component.ts`.

Related: #541, #546, #544

### DIFF
--- a/e2e/client-date-picker-revisit.spec.ts
+++ b/e2e/client-date-picker-revisit.spec.ts
@@ -145,7 +145,9 @@ test.describe('Client form date pickers', () => {
       await page.getByRole('button', { name: 'Next' }).click();
     };
 
-    await page.getByRole('link', { name: 'Clients' }).click();
+    // Reached by URL rather than through the sidebar, which is collapsed at phone width. The
+    // revisit that triggers the bug still happens in-app, via the list page's create button.
+    await page.goto('/clients');
     await expect(page).toHaveURL('/clients');
 
     // First client, created in full so the second visit starts from a torn-down form.
--- a/e2e/date-picker-revisit.spec.ts
+++ b/e2e/date-picker-revisit.spec.ts
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { test, expect } from './fixtures';
+
+const HEAD_OFFICE = 'Head Office';
+
+/**
+ * Regression cover for #541, rolled out across the routed forms.
+ *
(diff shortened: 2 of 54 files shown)


Aman-Mittal:
Pushed a fix for the failing **E2E (mobile viewport)** job, and rebased onto current `main`.

The failure was caused by this PR, but it surfaced in a spec this PR did not originally contain. `DUAL_VIEWPORT_SPECS` gains the glob `**/*date-picker-revisit.spec.ts`, which also matches `client-date-picker-revisit.spec.ts` — merged from #546 while this branch was open. That spec reached the client list by clicking the sidebar link, which is collapsed at phone width, so it timed out on the click as soon as the glob pulled it into the mobile project. Nothing about the component change was wrong; the new coverage just reached a spec that had only ever run on desktop.

It now navigates to the list by URL. The revisit that triggers the bug still happens in-app through the list's create button, so the spec covers the same failure at both viewports.

Verified on the rebased branch:

- `npx playwright test --project=mobile` — 19 passed, the whole mobile project including `mobile-shell` and `guidance-tour`
- `npx playwright test date-picker-revisit --project=mocked` — 3 passed

Before pushing I reproduced CI's exact tree locally (this branch merged with `main`) and confirmed the mobile job failed there for this reason, then that the same tree passes with the navigation change.

PR #556, opened 2026-09-17
fix(idle): dismiss inactivity dialog if session ended while it was pr…

### LINKED ISSUE #555: When Session is already timedout it still show keep me logged in dialog
### What happened

When Session is already timedout it still show keep me logged in dialog

### Steps to reproduce

Image

### Screen or route

_No response_

### Browser console and network errors

```text

```

### Which backend

Local docker stack (deploy/docker-compose-e2e.yml)

### Fineract version

_No response_

### Browser and version

_No response_

### Before submitting

- [x] I searched existing issues for this.
- [x] This is a bug in the UI, not in the Fineract platform.

Morrowga:
Fixes #555

## The bug

`IdleService.showInactivityWarning()` calls `dialogService.present()` and only assigns
`this.dialogRef` once that promise resolves. But `presentModal()` in
`ionic-overlay.adapter.ts` resolves only after Ionic's own enter animation finishes
(~300ms+, per that file's own comment)  not the instant the modal is created.

If the session ends for an unrelated reason during that window — a 401 from another tab,
an explicit sign-out `closeDialog()` runs while `dialogRef` is still `null` and has
nothing to dismiss. The warning dialog still appears moments later, even though the
session it's warning about is already gone.

## The fix

Check `authService.isAuthenticated()` at the first moment a real handle actually exists
(inside the `.then()` callback), and immediately dismiss the modal instead of treating it
as a live warning, when the session has already ended by that point.

## Testing

Added a test using the same deferred-promise pattern the existing
`'keeps the hard logout timer running when activity is reported while the warning is
still being presented'` test already uses to simulate this exact async gap  just
triggered by the session ending instead of an activity event.

Confirmed via `git stash` that the new test fails on the prior code with a precise
assertion failure (`expected "vi.fn()" to be called at least once`), not a type error —
and passes with the fix.

Full local check suite, all green:
- `npx tsc --noEmit`
- `npm run lint`
- `npm run format:check`
- `npm test -- --watch=false` 241/241 files, 1466/1466 tests
- `npm run check:icons`
- `npm run i18n:check`
- `npm run build`

## What I didn't do

This fixes the specific race described in #555. I did not audit `closeDialog()`'s other
callers or the broader `presentModal()` contract for similar async-gap issues elsewhere
in the app , this PR is scoped to the inactivity dialog only.


Aman-Mittal:
@Morrowga Pull request can be merged if the PR commits are not verified.

Aman-Mittal:
@Morrowga Are you working on this?

Morrowga:
> @Morrowga Are you working on this?

I am sorry I thought its already merged. I will do it.

Aman-Mittal (approved):
lgtm and thanks for contribution

